PR #243, opened 2015-01-21
Add angle vector compressed

furushchev:
this is pr2-interface whose angle-vector is sent as `uint8[17]` vector for decrease transfer data size and its bridge tool.


garaemon:
* decompressed joint states should be published rather than use it just inside of euslisp. You don't need to hack robot-interface. Please write more simple code.
* You can write a function independent from specific robots. Of course, you may need to have robot model to know min-max angles


Also please check following script.
https://github.com/start-jsk/rtmros_gazebo/tree/master/hrpsys_gazebo_atlas/euslisp/atlas-joint-state

jsk_pr2_robot/jsk_pr2_startup/jsk_pr2_teleop/pr2-compressed-angle-vector-client.l
```diff
@@ -0,0 +1,29 @@
...
+(ros::advertise "potentio_vector_compressed" jsk_pr2_startup::AngleVectorCompressed 1)
+
+(ros::rate 10)
+
+(defun publish-compressed-joint-state ()
+  (let ((msg (compress-angle-vector *pr2* (send *ri* :state :potentio-vector))))
+    (ros::ros-info "compressed msg: ~A" (send *ri* :state :potentio-vector))
+    (ros::publish "potentio_vector_compressed" msg)))
+
+(while (ros::ok)
```
garaemon: it looks that `ros::spin` is enough

jsk_pr2_robot/jsk_pr2_startup/jsk_pr2_teleop/pr2-compressed-angle-vector-interface.l
```diff
@@ -0,0 +1,68 @@
...
+   (av &optional tm &rest args)
+   (if tm (ros::ros-warn "time is ignored"))
+   (ros::publish "angle_vector_compressed" (compress-angle-vector robot av))
+   (send robot :angle-vector av))
+  (:angle-vector-sequence
+   (avs &optional tms &rest args)
+   (if tms (ros::ros-warn "time-lst is ignored"))
+   (dolist (av avs)
+     (send self :angle-vector av)
+     (unix:sleep 3)))
```
garaemon: sleep?

jsk_pr2_robot/jsk_pr2_startup/jsk_pr2_teleop/pr2-compressed-angle-vector-interface.l
```diff
@@ -0,0 +1,68 @@
...
+  (:state
+   (&rest args)
+   (case (car args)
+     (:potentio-vector
+      (send self :update-current-angle-vector)
+      (return-from :state current-angle-vector))
```
garaemon: no need to write `return-from`

jsk_pr2_robot/jsk_pr2_startup/jsk_pr2_teleop/pr2-compressed-angle-vector-interface.l
```diff
@@ -0,0 +1,68 @@
...
+     (:reference-vector
+      (send self :update-current-angle-vector)
+      (while updating)
+      (return-from :state current-angle-vector))
```
garaemon: no need to write `return-from`

garaemon:
I understood.
You implemented interface for compressed :potentio-vector and compressed :angle-vector in robot-interface.

About :angle-vector method, it whould be reasonable.
However, :potentio-vector method, I think the design is not best.
We should have simple compressor/decompresser nodes.

```
[/joint_states] --> (compressor) --> [compressed message] --> (decompressors) -> [/joint_states]
```

garaemon:
And also I think most of code can be (should be) re-usable across different robots, jsk_robot_startup package may be better.

PR #253, opened 2015-02-04
add self_filter launch and config to jsk_baxter

aginika:
Hand-made parameter tuning for padding.

Is there any way to publish colorized pointcloud?
self_filter_baxter

### DIFF
--- a/jsk_baxter_robot/jsk_baxter_startup/jsk_baxter_sensors/openni_self_filter.yaml
+++ b/jsk_baxter_robot/jsk_baxter_startup/jsk_baxter_sensors/openni_self_filter.yaml
@@ -1,0 +1,42 @@
+min_sensor_dist: .2
+self_see_default_padding: .01
+self_see_default_scale: 1.0
+self_see_links:
+  - name: right_gripper_base
+    padding: .04 
+  - name: right_lower_elbow
+    padding: .04 
+  - name: right_lower_forearm
+    padding: .04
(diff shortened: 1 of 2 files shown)


garaemon:
There shoud be parameter like use_color or somethin like that.
Please check it

garaemon:
These PRs

* https://github.com/PR2/pr2_navigation/pull/12
* https://github.com/PR2/pr2_navigation/pull/11

garaemon:
Please update it by the next PR

aginika:
Ok, Thanks!

garaemon:
And... Could you check PR2 uses organized filter?

aginika:
I'm afraid that I couldn't understand fully what you mean .

Now PR2 use pr2_navigation_self_filter only for tilt_laser and base_laser in pr2.launch.
However, there is a openni_cloud_self_filter_color.launch in jsk_pr2_startup, so we can use it to publish self_filter with just launching it.( need to modify a little for not using use_rgb and keep_organized)

I will confirm whether it works or not and update to use the use_rgb and keep_organized.

garaemon:
> However, there is a openni_cloud_self_filter_color.launch in jsk_pr2_startup, so we can use it to publish self_filter with just launching it.( need to modify a little for not using use_rgb and keep_organized)

Please add it to default, because the filter does not subscribe input if there is no subscriber of output topic.

PR #284, opened 2015-03-18
[baxtereus] overwrite baxter angle-vector for setting scale 5

aginika:
From jsk-ros-pkg/jsk_pr2eus#121


k-okada:
this is obsoleted PR, so close this

PR #289, opened 2015-04-03
- [baxter-interface.l] we found that input data must be larget then 3, and add dummy last element works very nice!

k-okada:


wkentaro:
I think this works. why don't you merge this?

k-okada:
sorry i Just forget that, thanks for ping me

PR #291, opened 2015-04-07
pepper-init added

kochigami:
I added "pepper-init" , which is similar to "nao-init", to make robot instances more easily.


k-okada:
nice!

PR #311, opened 2015-05-01
[baxtereus] add Baxter Safe Interface

aginika:
Thanks to @k-okada #303


k-okada:
how about `(baxter-init :safe t)` ?
there are no such convention, but i feel this type of usage is slightly 
major.

--
◉ Kei Okada

aginika:
@k-okada Thank you for your advise I updated

PR #314, opened 2015-05-02
- [baxter-util.l] comment out test code

k-okada:


k-okada:
#308 is wrong, if you load "baxter-util.l", new instance is created, which is not good.

PR #316, opened 2015-05-04
[roseus_remote] use ros::spin instead of ros::spin-once loop

furushchev:


k-okada:
MEMO: updating travis again due to https://github.com/jsk-ros-pkg/jsk_robot/pull/310, is there way to  check wrong commit of .travis?

PR #317, opened 2015-05-04
move baxter-robot-safe class to baxter-util.l, fix :self-collision-check and add test code

k-okada:
@aginika, please review this code and merge if this is ok

- [test/test-baxter.{l,test}] add test code for baxter model  (:self-collision-check)
- [baxter-util.l] use :collision-check-paris to get collision link pair,  instaed of combination
- [baxter-util.l] (length args) always retruns non nil, so it never goes to self-collision-check with pairs
- [baxter-interface.l, baxter-util.l] move baxter-robot-safe class definition to baxter-util.l


aginika:
@k-okada 
I executed and seems to be fine.

aginika:
Could you merge this?

PR #318, opened 2015-05-05
[baxtereus] overwrite baxter max joint velocity

aginika:
This will prevent baxter joint threshold exceeded better.


garaemon:
Report it to ML or baxter_description

PR #325, opened 2015-05-09
[baxtereus] fix baxter-interface :init args passing

aginika:

PR #329, opened 2015-05-13
[baxter-interface.l] Overwrite controller-type with default value in angle-vector-sequence

wkentaro:
Thanks to https://github.com/jsk-ros-pkg/jsk_robot/pull/325, by `(setq *ri* (instance baxter-interface :init :type :rarm-controller))` we can set default controller-type of the robot. (it will be set to the slots of *ri*)
But at this time, passing ctype to `:angle-vector` or `:angle-vector-sequence` raises a problem. It does not work when the passed ctype is different from default ctype.

I expect passed ctype is overwritten by default ctype (which is passed when initialization).

### DIFF
--- a/jsk_baxter_robot/baxtereus/baxter-interface.l
+++ b/jsk_baxter_robot/baxtereus/baxter-interface.l
@@ -244,6 +244,11 @@
 		 (when (= (length avs) 3) ;; when input avs is 1 or 2
 		   (setq avs (append avs (list (elt avs 2))))
 		   (setq tms (append tms (list 50))))
+     ;; place controller-type which was set when initialization above the passed argument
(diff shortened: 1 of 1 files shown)


wkentaro:
Just started reviewing :) <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

k-okada:
Sorry for late, what is the problem?
```
(load "package://baxtereus/baxter-interface.l")
(setq *ri* (instance baxter-interface :init :type :rarm-controller))
(send *ri* :angle-vector #f(0.0 20.0 -25.0 -40.0 60.0 -20.0 80.0 0.0 -20.0 -25.0 40.0 60.0 20.0 80.0 0.0))
```
works,

PR #332, opened 2015-05-15
use avoid-collision-distance to 5mm for baxter

k-okada:


k-okada:
this is rewrited version of https://github.com/start-jsk/2014-semi/pull/411

PR #345, opened 2015-06-09
Add english demo

kochigami:
@jiangjun0105 and I added English version demo to mirai-demo-20150413.l 
We can change pepper's demo according to language by executing
```demo :ja``` or ```demo``` (Japanese)
```demo :chi``` (Chinese)
```demo :en``` (English)


k-okada:
@kochigami @jiangjun0105 I think jsk_201504_miraikan package should move to 
under jsk_demo repository.

--
◉ Kei Okada

PR #357, opened 2015-06-24
modify speak method

kochigami:
I deleted "/" of speech topic in speak method 
in order to keep consistency with advertise topic "speech"

That's because when I executed ```send *ri* :speak "こんにちは"```, this error occurred.
```
[ERROR] [1435115396.024964819]: attempted to publish to topic /speech, which was not previously advertised. call (ros::advertise "/speech") first.
```
Sorry for not finding this error in the first commit.


garaemon:
I think you should modify `ros::advertise` instead of `ros::publish`

kochigami:
I added ```/``` to speech topic in both ```ros::advertise``` and ```ros::publish```.
Thank you very much for your advice.

PR #367, opened 2015-07-07
modify send-stiffness-controller method for pepper

kochigami:
According to http://doc.aldebaran.com/2-1/naoqi/motion/control-stiffness.html#control-stiffness,
stiffness of joints of the Pepper can be changed in only Head, LArm, RArm, LHand and RHand.
I wrote modified method of ```:send-stiffness-controller```   to change stiffness of one of these joints.
I checked all of them are changed by using this new method, but if there is a problem, please let me know.


k-okada:
lgtm

PR #371, opened 2015-07-16
.rosinstall remove pr2 deb files

k-okada:
can we remove pr2 repo from .rosinstall ?


k-okada:
if someone using jsk_robot/.rosinstall please let me know.

PR #373, opened 2015-07-17
[test/check_launch.test] add test code to check if pr2.launch is valid

k-okada:
check to wrong commit as #368

PR #375, opened 2015-07-17
[jsk_baxter_startup] add test_code to check if baxter.launch is valid

h-kamada:
like #373 , created baxter launch checker


k-okada:
taken over by #375

PR #377, opened 2015-07-17
[jsk_pepper_startup] add test_code to check if pepper launch is valid

h-kamada:
add test code for pepper launch


k-okada:
taken over by 380

PR #378, opened 2015-07-19
[jsk_pr2_startup/people_detection.launch] add people tracker

furushchev:
this PR enables people tracking using face/leg detection

### DIFF
--- a/jsk_pr2_robot/jsk_pr2_startup/jsk_pr2_sensors/lasers_and_filters.xml
+++ b/jsk_pr2_robot/jsk_pr2_startup/jsk_pr2_sensors/lasers_and_filters.xml
@@ -65,4 +65,11 @@
 
   </node>
 
+  <node name="base_scan_marking_to_laserscan"
+        pkg="pointcloud_to_laserscan" type="pointcloud_to_laserscan_node"
+        machine="c2" respawn="true" output="screen">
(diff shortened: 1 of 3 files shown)


k-okada:
are you sure that you did not updated package.xml to install leg_detector?
what is the role of scan filter?

furushchev:
@k-okada Thank you very much for confirming! Added deps to package.xml.
I added scan filter to get filtered data of `/base_scan` as `sensor_msgs/LaserScan` not as `PointCloud` in order to work with `leg_detector`

k-okada:
just to confirm
- 
https://github.com/jsk-ros-pkg/jsk_3rdparty/blob/master/jsk_ros_patch/laser_filters_jsk_patch/launch/footobject_filter_example.launch 
try to extract table foot, filtering "shaow" (noise) from lasers
- 
https://github.com/furushchev/jsk_robot/blob/add-people-tracker/jsk_pr2_robot/jsk_pr2_startup/jsk_pr2_sensors/lasers_and_filters.xml 
try to filter wall data ?

--
◉ Kei Okada

furushchev:
@k-okada Sorry for late response. I fixed to use existing filtered pointcloud data with conversion to LaserScan instead of creating it again from /base_scan.

PR #382, opened 2015-07-22
add sed command of correct spell

a-kako:
https://github.com/jsk-ros-pkg/jsk_robot/tree/master/jsk_pepper_robot

After setting up environment of Pepper according to this URL, error occured when loading pepper-interface.l. This is error message.

```
/opt/ros/indigo/share/euslisp/jskeus/eus/Linux64/bin/irteusgl 0 error: unbound \
variable pepper-robot in (defmethod pepper-robot (:reset-pose nil (send self :a\
ngle-vector #f(2.0 -2.0 -5.0 85.0 10.0 -70.0 -20.0 -40.0 85.0 -10.0 70.0 20.0 4\
0.0 0.0 0.0))))
```

I think it is due to invalid replacement of string in pepper.l.
This error has not occured after modifying CMakeLists.txt.


k-okada:
Thank you, we include this fix to #384 , with test codes

PR #385, opened 2015-07-22
change openni namespace to kinect_head

chiwunau:


k-okada:
btw, why we have kinect_head and kinect_head_c2 ?
from distributed object system philosophy, we  should use systems without
thinking about actual machine which the process is running...

--
◉ Kei Okada

garaemon:
pointcloud is too large to send over 1G ether. We should take care of communication of pointcloud seriously.

k-okada:
I mean why both topic publishes on same namespace, though it is send from
different computer?

--
◉ Kei Okada

garaemon:
same namespace?
There are kinect_head and kinect_head_c2

k-okada:
if the original  pipeline is
```
a/image -> a/image_throttle -> a/image_downsample -> a/results
```

and we split these process into several machines
```
(c0) -> a/image ->(c1) ->  a/image_throttle -> (c2) -> a/image_downsample
-> (c3) -> a/results
```
we do not have to change name space, but current program uses

```
(c0) -> a_c0/image ->(c1) ->  a_c1/image_throttle -> (c2) ->
a_c2/image_downsample -> (c3) -> a_c3/results
```

If changing namespace for each machine was wrong, we should fix that point
( https://github.com/jsk-ros-pkg/jsk_robot/pull/364/ ?)
and not to write patch depending on wrong implementation.


--
◉ Kei Okada

k-okada:
or if this is the limitation of nodelet or something, that's ok.

--
◉ Kei Okada

garaemon:
If we don't separate namespace, we cannot run pointcloud programs on different machines.

The programs run on c1 should listen to c1 pointcloud and the processes on c2 should listen to c2 pointcloud.

k-okada:
humm, so we do not have rule to which vision node run on c1 and which is on
c2, and all preprocessing pipeline will running on c1 and c2,


--
◉ Kei Okada

garaemon:
almost all the pipeline is run in c2 but driver is running on c1.

k-okada:
if so, there is a possibility to define rule that all vision process run on 
c2 and use same namespace for both machine.

https://github.com/jsk-ros-pkg/jsk_robot/pull/385#issuecomment-123940698

--
◉ Kei Okada

garaemon:
If vision processes run on c1, it would be ok.
But it's on c2 and everybody runs processes c1 first.
c2 suffix helps user to consider where vision processes should run.

2015年7月23日木曜日、Kei Okada<notifications@github.com>さんは書きました:

> if so, there is a possibility to define rule that all vision process run
> on
> c2 and use same namespace for both machine.
>
> https://github.com/jsk-ros-pkg/jsk_robot/pull/385#issuecomment-123940698
>
> --
> ◉ Kei Okada
>
>
> On Thu, Jul 23, 2015 at 1:47 PM, Ryohei Ueda <notifications@github.com
> <javascript:_e(%7B%7D,'cvml','notifications@github.com');>>
> wrote:
>
> > almost all the pipeline is run in c2 but driver is running on c1.
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <
> https://github.com/jsk-ros-pkg/jsk_robot/pull/385#issuecomment-123961767>
> > .
> >
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/jsk-ros-pkg/jsk_robot/pull/385#issuecomment-124088914>
> .
>


-- 
✉︎ from iPhone

PR #399, opened 2015-07-31
[jsk_robot_startup] Modify node name of gmapping and pointcloud_to_la…

orikuma:
test name is not appropriate


k-okada:
since I can not see log message, so not sure if this solves, but please look at #400,

PR #404, opened 2015-08-03
[jsk_robot_startup] Add rate param to modify tf publish rate and set …

orikuma:
…10.0 as defalut


k-okada:
Can you squash commits into one?

orikuma:
fixed

PR #405, opened 2015-08-04
[jsk_pr2_startup/jsk_pr2.rosinstall] add temporal missing package mongodb_store

furushchev:


k-okada:
lgtm, but dns at i is not working, if you need immediately, I'll merge this.

PR #419, opened 2015-08-12
[jsk_robot_startup] add jenkins/musca to database replication node

furushchev:


k-okada:
restarted travis, but why you need musca?

furushchev:
@k-okada thanks. I'd like to add musca to replication server because ISTCloud is more unstable than I thought. ( https://github.com/jsk-ros-pkg/jsk_robot/issues/407 )

k-okada:
I see, so robot-database is ist cloud machine, can you use Jenkins or pr2base machine for dis purpose?

furushchev:
@k-okada OK. I also added jenkins.  tested on PR1012.

Issue #359, opened 2015-06-24
`roslaunch jsk_pr2_startup pr2_hark.launch` failed with segmentation fault


furushchev:
below is log when executed `roslaunch --screen jsk_pr2_startup pr2_hark.launch` on PR1012:

```bash
UINodeRepository::Scan()
Scanning def /usr/lib/flowdesigner/toolbox
done loading def files
loading XML document from memory
node_LocalizeMUSIC_1 : param ELEVATION no longer used
node_GHDSS_1 : param TF_CONJ no longer used
node_GHDSS_1 : param MIC_FILENAME no longer used
node_GHDSS_1 : param MIC_POS_SHIFT no longer used
node_GHDSS_1 : param FIXED_NOISE no longer used
node_GHDSS_1 : param FIXED_NOISE_FILENAME no longer used
node_GHDSS_1 : param SPEED_OF_SOUND no longer used
node_GHDSS_1 : param UPDATE_ACCEPT_ANGLE no longer used
done!
Building network  :MAIN
whatisthis: transfer function
[E] [zip_open failed] [harkio_TransferFunction_fromFile] [153]
Segmentation fault
```

on PR1040:

```bash
Toolbox load error: libharkio3.so: cannot open shared object file: No such file or directory
Toolbox load error: libharkio3.so: cannot open shared object file: No such file or directory
There were errors loading the toolboxes:
/usr/lib/flowdesigner/toolbox/hark/hark.tlb
UINodeRepository::Scan()
Scanning def /usr/lib/flowdesigner/toolbox
done loading def files
loading XML document from memory
node_LocalizeMUSIC_1 : param ELEVATION no longer used
node_GHDSS_1 : param TF_CONJ no longer used
node_GHDSS_1 : param MIC_FILENAME no longer used
node_GHDSS_1 : param MIC_POS_SHIFT no longer used
node_GHDSS_1 : param FIXED_NOISE no longer used
node_GHDSS_1 : param FIXED_NOISE_FILENAME no longer used
node_GHDSS_1 : param SPEED_OF_SOUND no longer used
node_GHDSS_1 : param UPDATE_ACCEPT_ANGLE no longer used
done!
Building network  :MAIN
UINode.cc line 221: Node not found: MultiFFT
UINode.cc line 232: Exception caught while creating node_MultiFFT_1 (type MultiFFT)
UINetwork.cc line 559: Exception caught while building network LOOP0
[...]

PR #450, opened 2015-09-28
[jsk_pr2_startup/pr2_hark] fix issue https://github.com/jsk-ros-pkg/jsk_robot/issues/359

furushchev:
- Update Transfer function data for Hark >2.1
- Update batch files for new transfer function data
- Support stand-alone launch of `pr2_hark.launch`
- Update README

pr2_hark


furushchev:
**COMMENT**
It works, but still picks up much noise. So some another fix may be needed.

PR #513, opened 2015-12-07
[jsk_pr2_startup/jsk_pr2_move_base] split global_costmap_params for avoiding segfault

furushchev:
since recent change of indigo https://github.com/ros-planning/navigation/commit/8af3d5bd6722f257f356310b9860bb603ce488f0 , move_base_node segmentation fault on loading costmap2d::FootprintLayer plugin.
So I split our yaml file into two for hydro (=pr2 real robot) and for indigo.


furushchev:
please merge if there is no problem.

PR #515, opened 2015-12-08
[jsk_pr2_startup/pr2_gazebo.launch] fix: disable launch only for real robot; avoid check_cable on gazebo

furushchev:


k-okada:
humm, restart on travis did not work....

PR #518, opened 2015-12-09
fix leg end-coords for nao robot

k-okada:
-  add end-coords for nao robot
- naoqieus/cmake/compile_naoqi_model.cmake: .l depends on .yaml


k-okada:
before 
screenshot from 2015-12-09 18 44 59

aftter
screenshot from 2015-12-09 18 51 39

```
 (send-all (send *nao* :legs :end-coords) :draw-on :flush t :color #f(1 0 0))
```

k-okada:
cc: @ban-masa

PR #530, opened 2015-12-20
[jsk_pr2_startup/jsk_pr2_sensors/check_openni_node.py] refactor check_openni_node.py

furushchev:
- cleanup codes
- add `speak` parameter for enabling/disabling speaking feature
- add `sleep_cycle` for setting rate of check if openni node is alive
- add service advertise `~restart` to force invoke restarting openni node from service call


furushchev:
rebased origin/master

furushchev:
passed test.

PR #531, opened 2015-12-22
Remove unnecessary tf listener

orikuma:
Only removing tf broadcaster was not enough.


orikuma:
I will close this PR because this is included in #536

PR #532, opened 2015-12-23
Fix genjava problem caused by not listed message_generation

wkentaro:
Modified:
  - jsk_robot_common/roseus_remote/package.xml


wkentaro:
@k-okada
this is same problem as https://github.com/jsk-ros-pkg/jsk_visualization/pull/557.

PR #539, opened 2016-01-18
[jsk_pr2_startup/jsk_pr2_lifelog/objectdetection_db.py] unsubscribe objectdetection topics which has no other subscribers

furushchev:
This prevents to increase subscriber count of object detection nodes.

@h-kamada This may fix the problem of pointpose_extractor


furushchev:
I want to apply this fix to rosinstall for pr2 application.
@k-okada could you tag for this repo?

k-okada:
added tag 1.0.2

furushchev:
@k-okada Thank you very much!

PR #551, opened 2016-02-14
[jsk_pr2_startup/jsk_pr2.rosinstall] update jsk_robot to 1.0.2

furushchev:
for applying the fix in #539 
CC: @h-kamada 
Not yet be tested on PR2 real robot.


furushchev:
tested on PR1012.

PR #583, opened 2016-04-01
Convert to BGR image

wkentaro:
to handle mono and rgba images


wkentaro:
Please merge this, if it looks good!

PR #590, opened 2016-04-18
activate_user.l : fix for non pr2 robot, see #589

k-okada:
@furushchev please check if this works on pr2


furushchev:
added today's todo

furushchev:
@k-okada Sorry for very late. I confirmed on real robot. (PR1012)

k-okada:
Closes #589

PR #593, opened 2016-04-20
[baxtereus] Remove the loading baxter-util.l in baxter-interface.l

wkentaro:
In my case, I use baxter-interface.l but don't use baxter-util.l.

### DIFF
--- a/jsk_baxter_robot/baxtereus/baxter-interface.l
+++ b/jsk_baxter_robot/baxtereus/baxter-interface.l
@@ -1,5 +1,5 @@
 (load "package://pr2eus/robot-interface.l")
-(require :baxter "package://baxtereus//baxter-util.l")
+(require :baxter "package://baxtereus//baxter.l")
 (load "package://pr2eus/speak.l")
 (ros::load-ros-manifest "control_msgs")
 (ros::load-ros-manifest "baxter_core_msgs")


k-okada:
baxter-util.l re-define IK method, so robot behavior may change if you load baxter-util.l or not , I think this is very dangerous .
Could you explain why you do not need them?

--
◉ Kei Okada

2016/04/21 0:40、Kentaro Wada <notifications@github.com> のメッセージ:

wkentaro:
Reasons are:
1. Firstly, I think this kind of customization should be applied with user option, but currently user does not have option to use baxter-util or not.
2. Secondly, we use our own customized baxter model in our demo repository, it loads 1. baxter-interface.l, 2. customized baxter.l. So I'm scared about if the loading order is correct to overwrite the method (ex `:inverse-kinematics`) of baxter-robot class. (if 1. customized baxter.l 2. baxter-interface.l does not work, right?)

Is there best practice to use customized robot model with same robot interface?
I don't know much about `:baxter` in `(require :baxter "package://baxtereus/baxter.l")`, but does below work?
```
(require :baxter "customized-baxter.l")
(require "package://baxtereus/baxter-interface.l") ;; this also run (require :baxter "package://baxtereus/baxter-util.l")
```

k-okada:
A method defined in baxter-utils.l should be considered as a "default" method for the baxter robot, if you create your custom robot and override the ik, then re-define in your source code, if you want to change the IK behavior, change ik in baxter-utils, and if you want to debug IK, use something like ` (apply #'send-message self (class . super) :inverse-kinematics args)`
I think this is no longer needed as discussed in https://github.com/start-jsk/jsk_apc/pull/1365#event-646475953

wkentaro:
Yeah, thank you for checking our discussion.
I also reached the same conclusion with you that this PR is not required.

wkentaro:
I sent PR for chainging baxter-util.l. Could you please review it?
https://github.com/jsk-ros-pkg/jsk_robot/pull/602

PR #610, opened 2016-05-17
need to use collisions free-path when we send arbitary pose to :reset-pose in fetch robot

k-okada:
- feetcheus : fix typo in fetcheus.test
- fetcheus : add test for fetch-moveit-interface
- feteus-interface.l : always use moveit within :angle-vector


k-okada:
overtaken by #610

PR #615, opened 2016-05-24
[jsk_robot_startup] Define publisher before callback definition

orikuma:


k-okada:
why we need this?

orikuma:
Because publisher is used in callback functions and sometimes undefined error occurs in callbacks

PR #631, opened 2016-07-01
[jsk_robot_common/jsk_robot_startup/util/start_launch_sound.py] PLAY_START -> PLAY_ONCE

furushchev:
`PLAY_START` means starting repeating.
Current codes in `sound_play` does not support repeating, so it plays only once at now. 
I fixed the upstream, so this code should also be fixed.
https://github.com/ros-drivers/audio_common/pull/77


furushchev:
@k-okada Please merge if there is no problem.

PR #633, opened 2016-07-01
[jsk_robot_startup] Fix transformation from odom to map in slam_odom topic

orikuma:
mainly bug fix


orikuma:
This is bug fix, please merge

PR #635, opened 2016-07-11
adding new point just befor the final point produce jerkey motion

k-okada:
This will generate smoother motion, expecting not to produce jerkey motion specially the end of the trajectory
this code is introduced in https://github.com/jsk-ros-pkg/jsk_robot/pull/289, and @wkentaro said `I think this works. why don't you merge this?` so may be no longer need this code at some point...

Only tested on the gazebo, need to test on the real robot

PR #636, opened 2016-07-19
JTA insert current position to the first point,

k-okada:
this requries https://github.com/RethinkRobotics/baxter_interface/pull/73, if you run JTA with this feature, this procedure is done within joint_trajectory_action_server.py

see https://github.com/jsk-ros-pkg/jsk_robot/pull/635 for more detail


wkentaro:
Still does this PR need https://github.com/RethinkRobotics/baxter_interface/pull/73?

PR #632, opened 2016-07-01
[naoqieus] add set-external-collision-protection-status method

kochigami:
This is euslisp method for https://github.com/ros-naoqi/naoqi_bridge/pull/48
(enable/disable safety reflex of some body parts)

how to use
```
roslaunch jsk_pepper_startup jsk_pepper_startup.launch
roslaunch naoqi_apps externalCollisionAvoidance.launch nao_ip:=<Pepper's IP>
roseus pepper-interface.l
(send *ri* ::set-external-collision-protection-status 2 nil) ;; 2: Arms nil: disabled
```

related msg and srv files
https://github.com/ros-naoqi/naoqi_bridge_msgs/pull/6

(Note: for ```:get-external-collision-protection-status``` I'll change the pull-request of naoqi_bridge.)

PR #662, opened 2016-09-21
[naoqieus] use let for local variables in set(get)-external-collision-protection-status

kochigami:
This is a modification for the review in https://github.com/jsk-ros-pkg/jsk_robot/pull/632 .
Sorry for my mistake.


k-okada:
thank you

PR #669, opened 2016-09-28
[jsk_robot_startup] move initialpose_publisher.l from jsk_pr2_startup/jsk_pr2_move_base

furushchev:

PR #672, opened 2016-09-28
[jsk_robot_lifelog] move logging program from jsk_pr2_startup/jsk_pr2_lifelog to jsk_robot_startup/lifelog

furushchev:


furushchev:
Could you merge this if there is no problem?
Or should I put the pointer to the old directory for each logging program like https://github.com/jsk-ros-pkg/jsk_robot/pull/669?

k-okada:
are you sure that no one is using these files directory?

```
0  ...artup/jsk_pr2_lifelog/action_result_db.py → ...robot_startup/lifelog/action_result_db.py
File renamed without changes.
View 
0  ...2_startup/jsk_pr2_lifelog/move_base_db.py → ...jsk_robot_startup/lifelog/move_base_db.py
File renamed without changes.
View 
2  ...tup/jsk_pr2_lifelog/objectdetection_db.py → ...bot_startup/lifelog/objectdetection_db.py
```

furushchev:
@k-okada Not sure.
And in my opinion, I'd rather want to create new package (like a `jsk_robot_lifelog`) and put logging stuff into it (I should maintain!). They are not `startup` things any more and I want to extend with features (testing, simulating, analyzing...)

furushchev:
@k-okada How about reusing https://github.com/jsk-ros-pkg/jsk_db (I prefer renaming it to `jsk_log` or `jsk_lifelog`)? This seems to be inactive.

k-okada:
2016-10-12 22:05 GMT+09:00 Furushchev <notifications@github.com>:

> @k-okada <https://github.com/k-okada> Not sure.
>

please check that.

>  I'd rather want to create new package (like a jsk_robot_lifelog) and put
logging stuff into it (I should maintain!). They are not startup things any
more and I want to extend with features (testing, simulating, analyzing...)

I think it is better to put code with startup, because this means every
root need to be use when you start up, if you put these code within
different location or even different repository, not one will use them in
future, as we seen in jsk_db repo.


--
◉ Kei Okada

furushchev:
I think it is the same if `jsk_robot_startup` depends on the new package.
Or if it is ok on putting codes into `jsk_robot_startup`, it becomes very dirty. (It will includes gazebo files, mongodb scripts, converter scripts to `openease` and so on)

k-okada:
@furushchev please help @knorth55 and @kochigami to setup this feature on baxter and pepper

k-okada:
test

furushchev:
@k-okada Ok. Now I'm also working on refining this feature for collabolation with IAI system. Please wait for a while.

PR #679, opened 2016-10-08
add launch_joy in jsk_pepper_startup launch

knorth55:


k-okada:
please describe why you need this.

knorth55:
i just want to run pepper at 73B2 in full-time.
i didn't use joy stick to launch this file, and it shows error so that i add args.

kochigami:
Sorry for my late arrival, but 
it is also very useful when we use one interface which uses the joy_node not from jsk_pepper_startup, but from another package.  

I tried launchpad mini with my PC for Pepper.
When I pushed the button, I was surprised because Pepper suddenly said "インタラクタイブ" due to the joy_node from jsk_pepper_startup.
In addition, Pepper got ill because of interactive movement without purpose. :'(

PR #682, opened 2016-10-12
[jsk_naoqi_robot] add roseus in find_package

kochigami:
I'd like to have manifest.l of peppereus by just executing ```catkin b peppereus```. (reference: #678)

(Even though I can't test it on Nao (jsk_nao_startup) because Nao goes to travel now, I added same change like peppereus.)

If there is any problem, please let me know. Sorry for my trouble.


k-okada:
I see, you're right, this will solve https://github.com/jsk-ros-pkg/jsk_robot/issues/678

PR #691, opened 2016-10-26
add more admin docs

k-okada:
- add how to sync clocks (@wkentaro)

I'm not sure why we can not connect rviz this afternoon, but now it works ok after install chrony on my computer...not sure.


k-okada:
need to add eth0/wlan0 information on the robot @mmurooka

mmurooka:
can I disclose IP adress like this? or should not for security?
https://github.com/jsk-ros-pkg/jsk_robot/pull/691/files#diff-893e3cfcb6d06c862e2ead6fe88056dfR43

PR #698, opened 2016-11-01
more update on fetch-interface.l - set :use-torso as default behavior for :angle-vector - set base-controller-action-name nil

k-okada:


k-okada:
this requires https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/253

PR #699, opened 2016-11-02
add fetch-utils.l

k-okada:


k-okada:
this requires https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/255

PR #709, opened 2016-11-06
[fetcheus] [fetch-interface.l] tuckarm pose

furushchev:
This part of codes is rest on my repo without sending pull request.
If you don't use it, please just close without merge.


k-okada:
Since we are now uses moveit for :angle-vector, https://github.com/jsk-ros-pkg/jsk_robot/blob/1.0.7/jsk_fetch_robot/fetcheus/fetch-interface.l#L28
so we do not need any more. Is it correct?

furushchev:
@k-okada OK. So I close this time. Please use this if anyone needs this.

PR #716, opened 2016-11-09
[baxtereus] add moveit in baxter-interface

knorth55:
it succeeded in both simulation and real world. 
in simulation https://drive.google.com/open?id=0B5DV6gwLHtyJVEc4SnhxTjVzOG8
In real robot https://drive.google.com/open?id=0B5DV6gwLHtyJR0hpaDBoWVFMeE0

### DIFF
--- a/jsk_baxter_robot/baxtereus/baxter-interface.l
+++ b/jsk_baxter_robot/baxtereus/baxter-interface.l
@@ -1,4 +1,4 @@
-(load "package://pr2eus/robot-interface.l")
+(require "package://pr2eus_moveit/euslisp/robot-moveit.l")
 (require :baxter "package://baxtereus//baxter-util.l")
 (load "package://pr2eus/speak.l")
 (ros::load-ros-manifest "control_msgs")
@@ -11,7 +11,7 @@
 (defclass baxter-interface
   :super robot-interface
   :slots (gripper-sequence-id right-gripper-action left-gripper-action
-			      right-gripper-type left-gripper-type))
+			      right-gripper-type left-gripper-type moveit-robot))
 (defmethod baxter-interface
   (:init (&rest args)
    (prog1 (send-super* :init :robot baxter-robot :joint-states-topic "/robot/joint_states" :groupname "baxter_interface" args)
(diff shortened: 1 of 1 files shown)


jsk_baxter_robot/baxtereus/baxter-interface.l
```diff
@@ -9,9 +9,9 @@
 (defvar *wait-for-suction* 5000000)
 
 (defclass baxter-interface
-  :super robot-interface
+  :super robot-move-base-interface
```
k-okada: you do not have to use move-base-interface, that is for mobile robot

knorth55: thank you

knorth55:
fixed

k-okada:
do you need to override :angle-vector? if so we need to fix #719 too
https://github.com/jsk-ros-pkg/jsk_robot/blob/master/jsk_fetch_robot/fetcheus/fetch-interface.l#L26-L28

knorth55:
i read fetcheus and it does as below.

overriding `:angle-vector-motion-plan` -> `angle-vector`,
and renaming
`angle-vector` -> `angle-vector-raw`
`angle-vector-sequence` -> `angle-vector-sequence-raw`

i thought we need to be very careful to do so because exisiting programs may not work.
but, in future, `angle-vector-motion-plan` should be mainstream and `angle-vector-raw` be an option.
Therefore, i want to override the method.
@pazeshun what do you think?

pazeshun:
@knorth55
I agree with you. If we don't override, our followers may be confused.

knorth55:
Overriding methods #721

k-okada:
instead of changing code like
```
-  (send *ri* :angle-vector (send *baxter* :angle-vector) 10000)
 +  (send *ri* :angle-vector-raw (send *baxter* :angle-vector) 10000)
```
I think we should set default to nil for moveit feature and change
:angle-vector methods like
```
(:angle-vector (...)
  ;; for simulation mode   use :anlge-vector-raw
  (when (send self : simulation-modep)
   (return-from :angle-vector (send self :angle-vector-raw ...)))
 ;; when moveit is NOT enabeld, use :anlge-vector-raw
 (when (null (send self :set-moveit-environment)) ;; this just returns
moveit-evniroment
   (return-from :angle-vector (send self :angle-vector-raw ...)))
```




--
◉ Kei Okada


2016-11-11 21:55 GMT+09:00 Shingo Kitagawa <notifications@github.com>:

pazeshun:
+1 to @k-okada

PR #721, opened 2016-11-11
[baxtereus] override :angle-vector and rename existing methods

knorth55:
MERGE AFTER #719 
Before                       After
`:angle-vector`             -> `:angle-vector-raw`
`:angle-vector-sequence`    -> `:angle-vector-sequence-raw`
`:angle-vector-motion-plan` -> `:angle-vector`

related to https://github.com/jsk-ros-pkg/jsk_robot/pull/716#issuecomment-259861725 and https://github.com/jsk-ros-pkg/jsk_robot/pull/716#issuecomment-259869688

it works with simulation and real robot

### DIFF
--- a/jsk_baxter_robot/baxtereus/baxter-interface.l
+++ b/jsk_baxter_robot/baxtereus/baxter-interface.l
@@ -232,14 +232,32 @@
      (ros::publish "/robot/head/command_head_nod" msg)
      )
    )
-  (:angle-vector (av &optional (tm :fast) (ctype controller-type) (start-time 0) &rest args)
-		 (send* self :angle-vector-sequence (list av) (list tm) ctype start-time args))
-  (:angle-vector-sequence (avs &optional (tms :fast) (ctype controller-type) (start-time 0) &key (scale 2.2) (min-time 0.05))
+  (:angle-vector-raw (av &optional (tm :fast) (ctype controller-type) (start-time 0) &rest args)
+		 (send* self :angle-vector-sequence-raw (list av) (list tm) ctype start-time args))
+  (:angle-vector-sequence-raw (avs &optional (tms :fast) (ctype controller-type) (start-time 0) &key (scale 2.2) (min-time 0.05))
 		 ;; force add current position to the top of avs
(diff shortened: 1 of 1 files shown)


knorth55:
@pazeshun can you review this?

pazeshun:
If `mvit-rb` and `mvit-env` are nil, what happens?
I think it's better that we can use `:angle-vector` as `:angle-vetor-raw` in that situation.
@knorth55 What do you think?

knorth55:
> If mvit-rb and mvit-env are nil, what happens?

baxter-interface do not initialize moveit config.

https://github.com/knorth55/jsk_robot/blob/b8464bc2345a19a55e75c04dfb56b31c72286024/jsk_baxter_robot/baxtereus/baxter-interface.l#L49-L50

> I think it's better that we can use :angle-vector as :angle-vetor-raw in that situation.

method `:angle-vector` should be only for moveit, and in other case, robot should not move because it can be dangerous

pazeshun:
Ok, I understood.
+1

pazeshun:
https://github.com/jsk-ros-pkg/jsk_robot/pull/716#issuecomment-260166997

knorth55:
@pazeshun modified

knorth55:
modified

pazeshun:
+1

jsk_baxter_robot/baxtereus/baxter-interface.l
```diff
@@ -232,14 +232,32 @@
...
+- av : joint angle vector [rad]
+- tm : (time to goal in [msec]) ;; currently this value is ignored
+"
+   ;; for simulation mode
+   (when (send self :simulation-modep)
+     (return-from :angle-vector (send* self :angle-vector-raw av tm args)))
+   (when (not (numberp tm))
+     (ros::warn ":angle-vector tm is not a number, use :angle-vector av tm args"))
+   (if (and (get self :moveit-environment)
+            ((get self :moveit-environment) . robot))
```
knorth55: thank you!

knorth55:
fixed

PR #730, opened 2016-11-21
[baxtereus] add ctype in angle-vector methods for moveit

knorth55:
#### What changed

- add ctype in angle-vector for moveit
- add ctype in angle-vector-sequence for moveit
- use &key instead of &rest in baxter `angle-vector` methods


knorth55:
@k-okada  can you review and merge this?

PR #736, opened 2016-12-03
[jsk_pr2_startup] fix: pr2 move_base local cost map does not work on gazebo

furushchev:


furushchev:
@k-okada Please merge if there is no problem.

k-okada:
> pr2 move_base local cost map does not work on gazebo
because of indigo update? or never tested?

PR #741, opened 2016-12-05
[baxtereus] Allow initial av in angle-vector-sequence

### LINKED ISSUE #738: [Baxter] Baxter's arm goes slightly down when trajectory starts
In https://github.com/RethinkRobotics/baxter_interface/pull/74/files#diff-393fafc5e38a00ca79cd9d500e6ab03cR367 , joint current positions are set to first trajectory point. And those positions are got from `/robot/joint_states`.

But when Baxter's hand grasps heavy object (or gripper is heavy), `/robot/joint_states` and commanded positions are very different. In that situation, Baxter's arm goes slightly down because first trajectory point is different from current commanded joint angles.

pazeshun:
Requires https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/270
To fix #738

### DIFF
--- a/jsk_baxter_robot/baxtereus/baxter-interface.l
+++ b/jsk_baxter_robot/baxtereus/baxter-interface.l
@@ -237,10 +237,10 @@
     (send self :angle-vector-sequence-raw (list av) (list tm) ctype start-time :scale scale :min-time min-time))
   (:angle-vector-sequence-raw (avs &optional (tms :fast) (ctype controller-type) (start-time 0)
                                    &key (scale 2.2) (min-time 0.05) &allow-other-keys)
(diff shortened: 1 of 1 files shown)


pazeshun:
Tested in real robot.
I know why travis fails. It is because https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/270 is not released and `:angle-vector` doesn't have `allow-initial-av` key.

I don't sure about where to make this change, baxtereus or jsk_2016_01_baxter_apc.

pazeshun:
jsk-ros-pkg/jsk_robot#738 will be fixed in https://github.com/RethinkRobotics/baxter_interface/pull/77

PR #742, opened 2016-12-07
 [fetcheus] modify :reset-pose

708yamaguchi:
```
send *fetch* :torso :waist-z :joint-angle 20
```

moveit need more space between base and `elbow_flex_link`


knorth55:
you should change PR title for others like, 
- [fetcheus] modify :reset-pose

PR #746, opened 2016-12-16
[WIP] [baxtereus] Enable to set customized robot in baxter-interface

pazeshun:
### What I want to do
Set customized baxter-robot (jsk_2016_01_baxter_apc::baxter-robot) as `robot` in `baxter-interface`.
This enables us to use https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/244 without moving gripper joint.


pazeshun:
I realized that this is not needed. Sorry.

PR #758, opened 2017-03-22
Add base link frame and base odom topic param

orikuma:
For robots which base_link and odom topic name is different from BODY and /odom.
Default behavior is not changed (checked in JAXON)

PR #759, opened 2017-03-28
Add wheel odom publisher for common use and bicycle use

k-kimura:
With @orikuma vehicle SLAM demo, added wheel odometry publisher for common use.
Moreover, committed config and launch for bicycle odometry as wheel odometry sample.

Due to this change, enabled base_link_frame and base_odom_topic to be set as different values from default "BODY"and "/odom".

@orikuma Please check these commits.


k-kimura:
> enabled base_link_frame and base_odom_topic to be set as different values from default "BODY"and "/odom".

This change is related to #758.

k-kimura:
This is common util program for walking robots.
Please note that config parameters are required for each robot.

k-okada:
if so, please remove HRP2JSK_bicycle_odometry_params.yaml from here and put them in robot specific package.

k-kimura:
> if so, please remove HRP2JSK_bicycle_odometry_params.yaml from here and put them in robot specific package.

OK, I can move this yaml file from here to somewhere robot specific package.
However, there already seems to be other robot specific config files here.
https://github.com/jsk-ros-pkg/jsk_robot/tree/master/jsk_robot_common/jsk_robot_startup/config
So I think it is better to put this odometery yaml here, following the above.

k-kimura:
@orikuma san, Thank you for checking this PR.
I arranged changed files not to interfere with PR #758.

> I think currently wheel_odom_publisher is substantially specific for bicycle (it can be used in general nonholonomic 2d locomotion but currently only for wheel_odom_publisher_bicycle.launch)

I agree with this point.
Currently, I put wheel_odom_publisher_common.launch in common use with other vehicle tasks odometry.
But now this odometry publisher is used and tested only for bicycle task.
wheel_odom_publisher_bicycle.launch is one of the sample of vehicle odometry, and we can add and improve other wheel odometry programs in reference to this sample.

PR #768, opened 2017-04-24
[baxtereus] add wait key in :start-grasp

knorth55:
add `:wait` key as `:stop-grasp`

### DIFF
--- a/jsk_baxter_robot/baxtereus/baxter-interface.l
+++ b/jsk_baxter_robot/baxtereus/baxter-interface.l
@@ -130,8 +130,8 @@
    (send self :command-grasp baxter_core_msgs::EndEffectorCommand::*cmd_calibrate* arm :pos pos))
 
   (:start-grasp
-   (&optional (arm :arms) &key (effort 50))
-   (send self :go-grasp arm :pos 0 :effort effort)
+   (&optional (arm :arms) &key (effort 50) (wait nil))
+   (send self :go-grasp arm :pos 0 :effort effort :wait wait)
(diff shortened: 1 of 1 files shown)


k-okada:
on pr2, we always call with `:wait t`, and it is not configurable  https://github.com/jsk-ros-pkg/jsk_pr2eus/blob/0.3.10/pr2eus/pr2-interface.l#L244
on fetch, `:wait t` is default behavior, https://github.com/jsk-ros-pkg/jsk_robot/blob/1.0.9/jsk_fetch_robot/fetcheus/fetch-interface.l#L68

knorth55:
default value of `:wait`should be `t`?

knorth55:
Why `stop-grasp` only has `:wait` key word and default value is `nil`?
https://github.com/jsk-ros-pkg/jsk_pr2eus/blob/0.3.10/pr2eus/pr2-interface.l#L301

k-okada:
>Why stop-grasp only has :wait key word and default value is nil?

 not sure, but maybe 
```
(send *ri* :start-grasp)
(send *ri* :anlge-vector #f(...))
```
we have to wait until the robot actually grasp, when move arm to next position, but
```
(send *ri* :stop-grasp)
(send *ri* :anlge-vector #f(...))
```
the robot drop object just ofter :stop-grasp has called, and we can move arm, while the robot is trying to move gripper to initial position. May be because of speed up???

So we need to fix #769 ???

knorth55:
You mean `:stop-grasp` method should have default `:wait` value `t`?
Now I think default `:wait` value should be `t` in all gripper method.

PR #769, opened 2017-04-26
[baxtereus] set wait key default as t

knorth55:


knorth55:
https://github.com/jsk-ros-pkg/jsk_robot/pull/768#issuecomment-296886233

k-okada:
@knorth55 on dual arm manipulation case, you have to change your code when the robot grasp a object with two arms simultaneously

PR #774, opened 2017-04-28
Update jsk_fetch.rosinstall

Affonso-Gui:
Update version to master.

### DIFF
--- a/jsk_fetch_robot/jsk_fetch.rosinstall
+++ b/jsk_fetch_robot/jsk_fetch.rosinstall
@@ -6,4 +6,6 @@
     local-name: jsk_pr2eus
     uri: https://github.com/jsk-ros-pkg/jsk_pr2eus.git
     version: 0.3.7
-
+- git:
+    local-name: FA-I-sensor
+    uri: https://github.com/RoboticMaterials/FA-I-sensor.git
--- a/jsk_fetch_robot/jsk_fetch.rosinstall
+++ b/jsk_fetch_robot/jsk_fetch.rosinstall
@@ -1,9 +1,9 @@
 - git:
     local-name: jsk_robot
     uri: https://github.com/jsk-ros-pkg/jsk_robot.git
-    version: 824b165
+    version: master
 - git:
     local-name: jsk_pr2eus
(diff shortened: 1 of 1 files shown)


jsk_fetch_robot/jsk_fetch.rosinstall
```diff
@@ -1,9 +1,9 @@
 - git:
     local-name: jsk_robot
     uri: https://github.com/jsk-ros-pkg/jsk_robot.git
-    version: 824b165
+    version: master
 - git:
     local-name: jsk_pr2eus
     uri: https://github.com/jsk-ros-pkg/jsk_pr2eus.git
-    version: 0.3.7
+    version: master
```
k-okada: please specify confirmed version number

jsk_fetch_robot/jsk_fetch.rosinstall
```diff
@@ -1,9 +1,9 @@
 - git:
     local-name: jsk_robot
     uri: https://github.com/jsk-ros-pkg/jsk_robot.git
-    version: 824b165
+    version: master
```
k-okada: af1ad6919780042a5c6221b7aed4caaa88715267

jsk_fetch_robot/jsk_fetch.rosinstall
k-okada: https://github.com/jsk-ros-pkg/jsk_pr2eus/commits/master if there is important commit since 0.3.10, remove this line and use deb version of pr2eus

k-okada:
I think we need to cange
```
- git:
      local-name: jsk_robot
      uri: https://github.com/jsk-ros-pkg/jsk_robot.git
 -    version: 824b165
```
also we need to add approximate sensor package

Affonso-Gui:
Changed jsk_robot version to af1ad69  (after changed head end coords)

Will add Proximity sensor package later @708yamaguchi

k-okada:
@Affonso-Gui @708yamaguchi please add proximity sensor package

Affonso-Gui:
Added to jsk_fetch.rosinstall . Might work only with force_proximity_ros, but should be safer with FA-I-sensor package altogether.

Original package had no versions, so left as master.

PR #778, opened 2017-05-11
[fetcheus] disable gui in fetch-moveit.test

knorth55:
- use simulation.launch instead of playground.launch
- disable gui in fetch-moveit.test


knorth55:
Everytime, I run `catkin run_tests` it starts gazebo GUI.
I disable it, and use lighter world.

PR #791, opened 2017-06-19
[fetcheus] update angle-vector and add angle-vector-sequence

knorth55:
update `fetch` `:angle-vector` to follow `baxter` `:angle-vector` with `moveit`

### DIFF
--- a/jsk_fetch_robot/fetcheus/fetch-interface.l
+++ b/jsk_fetch_robot/fetcheus/fetch-interface.l
@@ -44,21 +44,24 @@
                :start-offset-time start-offset-time :clear-velocities clear-velocities
                :use-torso use-torso args))
   (:angle-vector-sequence
-   (avs &optional (tm 3000) (ctype controller-type) (start-time 0) &rest args
+   (avs &optional tms (ctype controller-type) (start-time 0) &rest args
         &key (use-torso t) (start-offset-time 0.01) (clear-velocities t) &allow-other-keys)
    "Send joind angle to robot with self-collision motion planning, this method returns immediately, so use :wait-interpolation to block until the motion stops.
 - avs : sequence of joint angle vector [rad]
-- tm : time to goal in [msec]
+- tms : list of time to goal from previous angle-vector point in [msec]
 - use-torso : set t to use torso
 "
    (setq ctype (or ctype controller-type))  ;; use default if ctype is nil
    ;; for simulation mode
    (when (send self :simulation-modep)
      (return-from :angle-vector-sequence
-                  (send* self :angle-vector-sequence-raw avs tm ctype start-time args)))
-   (when (not (numberp tm))
-     (ros::warn ":angle-vector tm is not a number, use :angle-vector av tm args"))
-   (send* self :angle-vector-motion-plan avs :ctype ctype :move-arm :rarm :total-time tm
+                  (send* self :angle-vector-sequence-raw avs tms ctype start-time args)))
+   (unless (and (listp tms) (every #'numberp tms))
+     (ros::warn ":angle-vector-sequence tms is not a list of number, use :angle-vector-sequence av tms args"))
+   (if tms
+     (setq tms (apply #'+ tms))
+     (setq tms 3000))
+   (send* self :angle-vector-motion-plan avs :ctype ctype :move-arm :rarm :total-time tms
                :start-offset-time start-offset-time :clear-velocities clear-velocities
                :use-torso use-torso args))
   (:default-controller ()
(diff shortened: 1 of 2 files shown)


k-okada:
could you explain whey we we need to update these method?

knorth55:
I set default values for `:angle-vector` method.
And fetch-users also want to use `:angle-vector-sequence` with `MoveIt!`

k-okada:
in that case, please consider putting these functions in robot-interface.l and fetch-interface.l can use these features with minimum changes.

knorth55:
this method is actually modified for `fetch`.
these cannot be applied to other robots.

k-okada:
it seems we have several improvement in this PR

1. eable to use angle-vector-sequence -> ok
2. use ctype?
3. enable to use angle-vector-raw if time is nil -> I think we will not support a way to call :angle-vector-raw implicitly for fetch robot.

knorth55:
> 2. use ctype?

these changes are needed to adapt to default `:angle-vector` method.
default `:angle-vector` pass `ctype` in `&optional`, but `:angle-vector-motion-plan` pass `ctype` in `&key`

> 3. enable to use angle-vector-raw if time is nil -> I think we will not support a way to call :angle-vector-raw implicitly for fetch robot.

if `tm` is `nil`, use `angle-vector-motion-plan` with `tm` `3000`

knorth55:
Should I split this commit into two?

knorth55:
Updated

jsk_fetch_robot/fetcheus/fetch-interface.l
```diff
@@ -25,24 +25,42 @@
...
-   (let ((use-torso t))
-     (if (and (member :use-torso args) (null (cadr (member :use-torso args)))) (setq use-torso nil))
-     (send self :angle-vector-motion-plan av :move-arm :rarm :total-time tm :start-offset-time 0.01 :use-torso use-torso :clear-velocities t)
-   ))
-  ;;
+   (send* self :angle-vector-motion-plan av :ctype ctype :move-arm :rarm :total-time tm
+               :start-offset-time start-offset-time :clear-velocities clear-velocities
+               :use-torso use-torso args))
+  (:angle-vector-sequence
+   (avs &optional (tm 3000) (ctype controller-type) (start-time 0) &rest args
```
k-okada: :angle-vector-sequence takes `tms`, list of time
https://github.com/jsk-ros-pkg/jsk_pr2eus/blob/0.3.10/pr2eus/robot-interface.l#L400

jsk_fetch_robot/fetcheus/fetch-interface.l
```diff
@@ -25,24 +25,42 @@
...
+        &key (use-torso t) (start-offset-time 0.01) (clear-velocities t) &allow-other-keys)
+   "Send joind angle to robot with self-collision motion planning, this method returns immediately, so use :wait-interpolation to block until the motion stops.
+- avs : sequence of joint angle vector [rad]
+- tm : time to goal in [msec]
```
k-okada: `tms: list of time to goal from previous angle-vector point in [msec]`?

jsk_fetch_robot/fetcheus/fetch-interface.l
```diff
@@ -25,24 +25,42 @@
...
+- use-torso : set t to use torso
+"
+   (setq ctype (or ctype controller-type))  ;; use default if ctype is nil
+   ;; for simulation mode
+   (when (send self :simulation-modep)
+     (return-from :angle-vector-sequence
+                  (send* self :angle-vector-sequence-raw avs tm ctype start-time args)))
+   (when (not (numberp tm))
+     (ros::warn ":angle-vector tm is not a number, use :angle-vector av tm args"))
+   (send* self :angle-vector-motion-plan avs :ctype ctype :move-arm :rarm :total-time tm
```
k-okada: `apply #'+ tms` ?

jsk_fetch_robot/fetcheus/fetch-interface.l
```diff
@@ -25,24 +25,42 @@
...
+- avs : sequence of joint angle vector [rad]
+- tm : time to goal in [msec]
+- use-torso : set t to use torso
+"
+   (setq ctype (or ctype controller-type))  ;; use default if ctype is nil
+   ;; for simulation mode
+   (when (send self :simulation-modep)
+     (return-from :angle-vector-sequence
+                  (send* self :angle-vector-sequence-raw avs tm ctype start-time args)))
+   (when (not (numberp tm))
```
k-okada: `every #'numberp tms` ?

jsk_fetch_robot/fetcheus/fetch-interface.l
knorth55: Should I modify `pr2eus_moveit` `:angle-vector-motion-plan` to support `tms` as list of time?

knorth55:
Updated

PR #790, opened 2017-06-17
Refactoring lifelog nodes

furushchev:
This PR is part of https://github.com/jsk-ros-pkg/jsk_robot/pull/696

- Enable test for logging nodes (using gazebo)
- Use TF2 to resolve frames (fallback TF1 on failure)
- Enable non-blocking insertion to database (needs https://github.com/strands-project/mongodb_store/pull/184, release requested https://github.com/strands-project/mongodb_store/issues/187 )
- Update launch files (for PR2 / Fetch)

PR #822, opened 2017-08-24
enable to pass travis on kinetic

k-okada:
- remove CMakeLists.txt: pr2_* from CATKIN_DEPENDS
- add .travis.rosinstall.kinetic

### DIFF
--- a/.travis.rosinstall
+++ b/.travis.rosinstall
@@ -1,29 +1,26 @@
-- git:
-   uri: http://github.com/RethinkRobotics/baxter_common.git
-   local-name: RethinkRobotics/baxter_common
-- git:
-    uri: https://github.com/jsk-ros-pkg/jsk_demos.git
-    local-name: jsk-ros-pkg/jsk_demos
-- git:
-    uri: https://github.com/jsk-ros-pkg/jsk_smart_apps.git
-    local-name: jsk-ros-pkg/jsk_smart_apps
-- git:
-    uri: https://github.com/jsk-ros-pkg/openrave_planning.git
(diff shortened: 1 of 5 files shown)


.travis.rosinstall
```diff
@@ -1,29 +1,26 @@
...
-- git:
-    uri: https://github.com/jsk-ros-pkg/jsk_smart_apps.git
-    local-name: jsk-ros-pkg/jsk_smart_apps
-- git:
-    uri: https://github.com/jsk-ros-pkg/openrave_planning.git
-    local-name: jsk-ros-pkg/openrave_planning
-- git:
-    uri: https://github.com/code-iai/SnapMapICP.git
-    local-name: tum-ros-pkg/highlevel/SnapMapICP
+# - git:
```
furushchev: @k-okada This should not be removed but be moved to `.traivs.rosintall.hydro` until PR1040 is upgraded?

furushchev: Note: I'm working on another pull request and trying to test pr2 stuff both on hydro (from source) and on indigo (from apt).
https://github.com/jsk-ros-pkg/jsk_robot/pull/790/files#diff-354f30a63fb0907d4ad57269548329e3L26

k-okada: I found we do not need this for travis test. So If we need this, may be add them to pr2 related packages.

furushchev: @k-okada OK. moved to `.travis.rosinstall.hydro` and set up `.travis.yml` to test only pr2 related packages on hydro.
(Actually other packages like drc related packages always fails at this momemt.)

PR #824, opened 2017-08-24
test/test-baxter-moveit.l: sleep with ros::time, not unix::time

k-okada:
using unix::sleep, sleeps against wall clock, we'd like to use ros::time for travis test


k-okada:
@k-okada Not sure why you added ''wait for upstream" tag????? once roseus 1.7.0 released, we can remove this code https://github.com/ros/rosdistro/pull/18436,

PR #833, opened 2017-09-06
jsk_fetch_startup: add voice_text and speak-jp

furushchev:
- Installed voice text to fetch15.
- Added text to speech nodes to bringup launch file
- Moved `:speak` methods to robot interface (require https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/318 )

### DIFF
--- a/jsk_fetch_robot/fetcheus/fetch-interface.l
+++ b/jsk_fetch_robot/fetcheus/fetch-interface.l
@@ -130,9 +130,15 @@
       (setq result (send gripper-action :get-result))
       result))
   ;;
-  (:speak (text &rest args)
-    (let ()
-      (apply #'speak-en text :topic-name "/sound_play" args)))
+  (:speak-en (text &rest args)
+    (apply #'speak-en text :topic-name "/sound_play" args))
+  (:speak-jp (text &rest args)
+    (apply #'speak-jp text :topic-name "/robotsound_jp" args))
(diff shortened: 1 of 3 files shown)


jsk_fetch_robot/fetcheus/fetch-interface.l
```diff
@@ -130,9 +130,15 @@
       (setq result (send gripper-action :get-result))
       result))
   ;;
-  (:speak (text &rest args)
-    (let ()
-      (apply #'speak-en text :topic-name "/sound_play" args)))
+  (:speak-en (text &rest args)
```
k-okada: is this common interface with pr2 robot or robot-interface?
Cc: @snozawa

furushchev: @k-okada This is not common now but uses the same functions as PR2. AFAIK, this style is first emerged in `naoqieus` https://github.com/jsk-ros-pkg/jsk_robot/blob/master/jsk_naoqi_robot/naoqieus/naoqi-interface.l#L250 .

k-okada:
so, I think we're better to add this method to robot-interface.l

--
◉ Kei Okada


2017-09-07 22:19 GMT+09:00 Furushchev <notifications@github.com>:

furushchev:
@k-okada Thanks for advice! I created PR https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/318 to move these methods and removed `:speak` from fetch-interface instead.

PR #834, opened 2017-09-06
Add throttled RGB / Pointcloud topics for RViZ

furushchev:
deriving #833 

fetch_rviz

PR #850, opened 2017-10-05
travis.yaml: set ROS_PARALLEL_TEST_JOBS to 1

k-okada:
running two gazebo instance on test???

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ env:
     - USE_DOCKER=true
     - ROS_PARALLEL_JOBS="-j2"
     - CATKIN_PARALLEL_JOBS="-p2"
-    - ROS_PARALLEL_TEST_JOBS="-j4"
+    - ROS_PARALLEL_TEST_JOBS="-j1"
     - CATKIN_PARALLEL_TEST_JOBS="-p1"
     - NOT_TEST_INSTALL=true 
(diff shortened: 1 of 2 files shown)


furushchev:
@k-okada I tried the same thing and it seems that is not the only reason...
https://github.com/jsk-ros-pkg/jsk_robot/pull/834
https://travis-ci.org/jsk-ros-pkg/jsk_robot/jobs/283088941

k-okada:
but my PullRequest passed without restart.....

furushchev:
@k-okada Can I restart this test? I saw sometimes it passes and sometimes not.

k-okada:
yes, but I'm suspect if you restart, it fails. and if I restart, it passes. :-)

furushchev:
@k-okada Now, you might be correct ;-)

k-okada:
@furushchev ok, i have added new commit and passed, can you restart that, to check if that REALLY works ;-)

furushchev:
@k-okada Okay, this is the only thing only I can do! If it now passes, it means virtual display breaks moveit (or moveit with gazebo)?

furushchev:
@k-okada I restarted and now it passed again!

furushchev:
@k-okada Thanks a lot for debugging! :+

PR #852, opened 2017-10-17
Add use-base key in fetch-utils inverse-kinematics

Affonso-Gui:
Don't know if should be useful or not, but tried adding :use-base option for fetch's :inverse-kinematics, according to PR2's example [pr2-utils](https://github.com/jsk-ros-pkg/jsk_pr2eus/blob/master/pr2eus/pr2-utils.l/#L106-L151).

Code can be verified with
```
(send *fetch* :inverse-kinematics (make-coords :pos #f(1150 0 786)) :use-base t)
```
or
```
(send *fetch* :inverse-kinematics (make-coords :pos #f(1200 0 786)) :use-base t :base-range (list :min #f(0 0 0) :max #f(100 100 0)))
```

### DIFF
--- a/jsk_fetch_robot/fetcheus/fetch-utils.l
+++ b/jsk_fetch_robot/fetcheus/fetch-utils.l
@@ -5,14 +5,57 @@
 
 (defmethod fetch-robot
   (:inverse-kinematics
-   (target-coords &rest args &key link-list move-arm (use-torso t) move-target &allow-other-keys)
-   (unless move-arm (setq move-arm :rarm))
-   (unless move-target (setq move-target (send self :rarm :end-coords)))
-   (unless link-list
-     (setq link-list (send self :link-list (send move-target :parent)
-                           (unless use-torso (car (send self :rarm))))))
-   (send-super* :inverse-kinematics target-coords
-                :move-target move-target
-                :link-list link-list
-                args)
-   ))
+   (target-coords &rest args &key
+		  link-list move-arm (use-torso t) move-target
+                  (use-base nil) (start-coords (send self :copy-worldcoords))
+		  (base-range (list :min #f(-30 -30 -30)
+				    :max #f( 30  30  30)))
+		  additional-weight-list
+		  &allow-other-keys)
+  (let (diff-pos-rot)
+    (unless move-arm (setq move-arm :rarm))
+    (unless move-target (setq move-target (send self :rarm :end-coords)))
+    (unless link-list
+      (setq link-list (send self :link-list (send move-target :parent)
+			    (unless use-torso (car (send self :rarm))))))
+
+    (cond
+      (use-base
+       (setq diff-pos-rot
+             (concatenate float-vector
+                          (send start-coords :difference-position self)
+                          (send start-coords :difference-rotation self)))
+       (send self :move-to start-coords :world)
+       (with-append-root-joint
+	   (ll self link-list
+	       :joint-class omniwheel-joint
+	       :joint-args base-range)
+	 (send (caar ll) :joint :joint-angle
+	       (float-vector (elt diff-pos-rot 0)
+			     (elt diff-pos-rot 1)
+			     (rad2deg (elt diff-pos-rot 5))))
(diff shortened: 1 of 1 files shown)


k-okada:
I think @furushchev is working on this issues, specially from `move_base` side, please discuss with him.

Affonso-Gui:
What do you think about it, @furushchev ?

Affonso-Gui:
Since fetch uses `x` and `θ` for moving, above `:use-base` only works for `x` direction. Trying to move the robot in `y` direction leads to IK error, as if `:use-base` was set to `nil`.

Not very useful method, and does not behave as is PR2. Closing.

furushchev:
@Affonso-Gui @k-okada Sorry, I overlooked this issue.
Since fetch robot is non-holonomic, motion planning with base is not easy.
I recently implemented simple base action without navigation: https://github.com/jsk-ros-pkg/jsk_robot/blob/master/jsk_fetch_robot/fetcheus/fetch-interface.l#L233-L315
If it is inevitable to solve ik with base movement, it might be a workaround to combine ik and this base action, though it is far from elegant:
For example, when opening a large door

1. locate the robot
0. grasp the door handle
1. solve ik for opening 30 degrees
2. detach from the door handle
3. move base without navigation
4. repeat 0-3 until open the door

Please also note that moving non-holonomic base may cause huge base rotation even for very near destination, so please be careful not to break arm by colliding with objects around the robot.
Hope it helps.

Affonso-Gui:
@furushchev Thanks! Will take in consideration when needed

k-okada:
fetch is not omnidirectional, so use `wheel-joint` instead. see https://github.com/k-okada/jsk_robot/commit/4a208f77b779a5fb0991f7d251409d6df8a36911 for fix.
see
```
(progn
   (load "fetch-utils.l")
   (send *fetch* :reset-pose)(send *fetch* :newcoords (make-coords))
   (send *fetch* :inverse-kinematics (make-coords :pos #f(400 800 800) :rpy (float-vector 3 0 0)) :debug-view :no-message :use-base t :p-limit 10 :r-limit 0.05 :stop 500))
```
for example.


to show trajectory, see following example,
```
(defmethod wheel-joint
  (:joint-angle
   (&optional v &key relative &allow-other-keys)
   "return joint-angle if v is not set, if v is given, set joint angle. v is  joint-angle vector, which is  (floa\
t-vector translation-x[mm] rotation-z[deg])"
   (let (relvel relang)
     (unless relative
       (if v (warn "wheel-joint does not support non-relative mode??~%"))
       (return-from :joint-angle joint-angle))
     (when v
       (setq relvel (elt v 0) relang (elt v 1))
       (send child-link :translate (float-vector relvel 0 0))
       (send child-link :rotate (deg2rad relang) :z)
       (setq joint-angle (float-vector relvel relang))
       )
     joint-angle))
  )
(progn
  (load "fetch-utils.l")
  (send *fetch* :reset-pose)
  (send *fetch* :newcoords (make-coords))
  (send *fetch* :inverse-kinematics (make-coords :pos #f(400 800 800) :rpy (float-vector 3 0 0))
        :debug-view :no-message :use-base t :p-limit 10 :r-limit 0.05 :stop 500
         ;; to show velocities need patch to wheel-joint                                                          
        :move-joints-hook #'(lambda () (print (send *fetch* :links :joint :joint-angle))))
  )

(progn
  (load "fetch-utils.l")
  (send *fetch* :reset-pose)
  (send *fetch* :newcoords (make-coords))
  (send *fetch* :inverse-kinematics (make-coords :pos #f(400 800 800) :rpy (float-vector 3 0 0))
        :debug-view :no-message :use-base t :p-limit 10 :r-limit 0.05 :stop 500
        ;; display x-y, degree                                                                                    
        :move-joints-hook #'(lambda () (print (list (subseq (send *fetch* :worldpos) 0 2)
                                                    (rad2deg (elt (car (rpy-angle (send *fetch* :worldrot))) 0))
                                                    (send *fetch* :angle-vector)))))
  )
```

PR #857, opened 2017-11-01
travis: drop hydro

furushchev:


furushchev:
same situation as https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/328
To me, it's ok to be merged.

travis_robot

PR #870, opened 2017-11-21
Add :gripper method to fetch-interface

Affonso-Gui:
[robot-interface.l](https://github.com/jsk-ros-pkg/jsk_pr2eus/blob/master/pr2eus/robot-interface.l#L1038-L1042) states clearly that `:gripper` method should be defined in child classes, but `fetch-interface` had none defined.

Wrote based on `pr2-interface.l`, with the difference that it doesn't take `arm` argument, since fetch only has `:rarm`.

**Known problem:** fetch's gazebo simulation does not support /gripper_state topic. Currently inquiring fetch-robotics about this.

@708yamaguchi

### DIFF
--- a/jsk_fetch_robot/fetcheus/CMakeLists.txt
+++ b/jsk_fetch_robot/fetcheus/CMakeLists.txt
@@ -8,10 +8,8 @@ if($ENV{ROS_DISTRO} STREQUAL "hydro")
 endif()
 
 find_package(catkin REQUIRED
-    collada_urdf
-    euscollada
     fetch_description
-    roseus
+    roseus # required in indigo
 )
 
 catkin_package()
--- a/jsk_fetch_robot/fetcheus/CMakeLists.txt
+++ b/jsk_fetch_robot/fetcheus/CMakeLists.txt
@@ -7,7 +7,12 @@ if($ENV{ROS_DISTRO} STREQUAL "hydro")
   return()
 endif()
 
-find_package(catkin REQUIRED fetch_description)
+find_package(catkin REQUIRED
+    collada_urdf
+    euscollada
+    fetch_description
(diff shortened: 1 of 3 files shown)


708yamaguchi:
You may need to write below
```
<build_depend>fetch_driver_msgs</build_depend>
<run_depend>fetch_driver_msgs</run_depend>
```

708yamaguchi:
in `package.xml`

knorth55:
I add to use `ros::load-ros-manifest` and remove `(ros::roseus-add-msgs "fetch_driver_msgs")`.
Please check it.
https://github.com/Affonso-Gui/jsk_robot/pull/1

Affonso-Gui:
Thanks!

Affonso-Gui:
@knorth55 Changes worked just fine after rebuilding `fetcheus`. Thanks once again.

jsk_fetch_robot/fetcheus/CMakeLists.txt
```diff
@@ -7,7 +7,12 @@ if($ENV{ROS_DISTRO} STREQUAL "hydro")
   return()
 endif()
 
-find_package(catkin REQUIRED fetch_description)
+find_package(catkin REQUIRED
+    collada_urdf
+    euscollada
```
k-okada: If cmake does not use variables defined in collada_urdf and euscollada, we do not need that.


I think roseus is requred to compile fetch_driver_msgs, but it from kinetc, you do not need that, so please add some comment like `;; <= indigo required this,`

Affonso-Gui: @k-okada Fixed. Please correct me if I got it wrong.

PR #878, opened 2017-12-13
jsk_pr2_desktop: add autorestart

furushchev:
Add script to respawn application monitoring ros master process.


k-okada:
this is nice feature, but I have heard similar idea at somewhere, and you're not only one who implemented this kind of program, so if you find someone write similar code, please remember to update this.

mmurooka:
Maybe the following is similar feature.
https://github.com/jsk-ros-pkg/jsk_common/blob/master/jsk_tools/bin/roscore_regardless.py
https://github.com/start-jsk/rtmros_hrp2/blob/master/jsk_hrp2_ros_bridge/scripts/hrp2_stat.sh#L13

PR #918, opened 2018-04-23
fetch: enable replication

furushchev:
Without this, database size in fetch robot never decrease.


furushchev:
@k-okada 
Running on real robot one week without no problem.
Could you merge this?

PR #920, opened 2018-04-26
jsk_pr2_startup: specify prefix for laser_filters plugins

furushchev:
According to https://github.com/PR2/pr2_navigation/pull/27


k-okada:
So, we did not use shadow filters until now?

k-okada:
does this mean , if we https://github.com/PR2/pr2_navigation/pull/27, we have to update our file?
or both https://github.com/PR2/pr2_navigation/pull/27 and jsk_robot has wrong settings for laser filter and they are not working correctly right now. ?

PR #921, opened 2018-04-26
Run baxter on Kinetic

knorth55:


mmurooka:
default baxter experiment setting

kasihwa_baxter_dgx

PR #927, opened 2018-05-08
jsk_pr2_startup: add test for launch files

furushchev:
Added `roslaunch-check` for main launch files: `pr2_bringup.launch`, `pr2.launch` and `pr2_gazebo.launch`.
Fixed to pass tests.

### DIFF
--- a/jsk_pr2_robot/jsk_pr2_startup/CMakeLists.txt
+++ b/jsk_pr2_robot/jsk_pr2_startup/CMakeLists.txt
@@ -20,3 +20,14 @@ endforeach()
 
 install(FILES install_pr1040_description.sh jsk_pr2.machine plugin.xml startup.app
   DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION})
+
+if(CATKIN_ENABLE_TESTING)
+  find_package(rostest REQUIRED)
+  find_package(roslaunch REQUIRED)
+  set(roslaunch_check_script ${PROJECT_SOURCE_DIR}/test/roslaunch-check)
+  roslaunch_add_file_check(pr2_bringup.launch ROS_ENV_LOADER=/opt/ros/$ENV{ROS_DISTRO}/env.sh)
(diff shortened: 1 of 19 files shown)


k-okada:
Thank you for contribution with nice test code,

We're using ROBOT env about 15 years ago, but turned out that we'd remove number of required env variable as much as possible, 
For example, ROS depends on ROS_PACKAGE_PATH and a lot of people forget/confused about that
https://www.google.co.jp/search?q=ros_package_path+site%3Aanswers.ros.org

This is why I proposed sing rosparam for getting robot name and types, https://github.com/ros-infrastructure/rep/pull/104, Please consider using that

Since you're using  jsk_robot package, we set these params by ourselves,
https://github.com/jsk-ros-pkg/jsk_robot/blob/master/jsk_fetch_robot/jsk_fetch_startup/launch/fetch_bringup.launch#L10-L11
and created PR to upstream https://github.com/fetchrobotics/fetch_robots/pull/22

furushchev:
@k-okada I see. One problem I'd like to discuss after reading your comment is that `roslaunch` parse `args` only **before** putting all values into parameter servers.
In this case, we, PR2 users, use the `ROBOT` environment variable for switching machine definition to be loaded in launch files on launch time.
https://github.com/jsk-ros-pkg/jsk_robot/blob/master/jsk_pr2_robot/jsk_pr2_startup/pr2.launch#L31
Do you have any idea / direction on addressing this problem?

And I rather sent this pull request where we can use `ROBOT` as an **argument** instead of env variable because I don't like this confusing environment variable `ROBOT`.

k-okada:
https://github.com/jsk-ros-pkg/jsk_robot/blob/master/jsk_pr2_robot/jsk_pr2_startup/pr2.launch#L31
this just switches pr2/sim and usually top level launch file differs, such
as roslaunch pr2.launch or roslaunch sim launch

--
◉ Kei Okada


2018-05-10 13:01 GMT+09:00 Yuki Furuta <notifications@github.com>:

PR #919, opened 2018-04-25
lifelog: support persistent data in log

furushchev:
### Note
- Requires https://github.com/strands-project/mongodb_store/pull/220
- Resolve issue reported at https://github.com/jsk-ros-pkg/jsk_robot/pull/917#issuecomment-383814715

### Content

- Support persistent data on logging
  - Persistent data is defined as data with  `{'persistent': True}` in meta info
- Added an option to keep persistent data after replication by default
- Record base trajectory as persistent data by default

PR #928, opened 2018-05-08
jsk_robot_startup: use synchronized_throttle for camera

furushchev:
Merge this after #919 
Replaces `jsk_topic_tools/LightweightThrottle` with `jsk_topic_tools/SynchronizedThrottle`.

### DIFF
--- a/jsk_robot_common/jsk_robot_startup/lifelog/common_logger.launch
+++ b/jsk_robot_common/jsk_robot_startup/lifelog/common_logger.launch
@@ -50,32 +50,77 @@
         output="screen" respawn="$(arg respawn)"/>
 
   <group ns="lifelog">
+    <!-- image throttle -->
     <group if="$(arg save_rgb)">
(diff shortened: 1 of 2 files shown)


k-okada:
please update package.xml something like `<depend gte="1.x.x">jsk_topic_tools</depend>`

furushchev:
@k-okada Added `gte` attribute to `jsk_topic`tools` .
Since https://github.com/jsk-ros-pkg/jsk_common/pull/1579 is not yet released, I incremented the minor version from the current version.

k-okada:
```
- The "run_depend" tag must not have the following attributes: gte
```
please read my comment very carefully , 'something like' means the information is not correct. may be it should be "something like" 'version_gte'
http://www.ros.org/reps/rep-0127.html#build-depend-multiple

furushchev:
@k-okada Sorry, updated.

PR #936, opened 2018-06-04
[jsk_nao_startup] use common packages as jsk_pepper_startup

kochigami:
I'd like to use common packages as for jsk_nao/pepper_startup.

- quit using nao_interaction_launchers
- add launch_twitter param name

What I will do in a few days:
- add joy package


kochigami:
- added joy package
(+ set small value to scale_linear, scale_linear_turbo compared with pepper) 

memo:
Sometimes NAO continued walking. It was dangerous.
I noticed that NAO's walking stops if I pushed a left trigger button again.
=> I added a comment about a stop button.

PR #949, opened 2018-07-13
[jsk_naoqi_robot] add ros::sleep to speaking node

kochigami:
After this change, the CPU temperature of a PC which naoqi-ros is working becomes around 50.
(Before this change, it becomes over 80...)
I executed `roslaunch jsk_nao(pepper)_startup jsk_nao(pepper)_startup.launch`
I used this command to check the temperature: `cat /sys/class/thermal/thermal_zone0/temp`

Note:
After the node is killed, NAO speaks good night, but Pepper doesn't.
I don't know why. I couldn't have solved the issue yet.

### DIFF
--- a/jsk_naoqi_robot/jsk_nao_startup/launch/jsk_nao_startup.launch
+++ b/jsk_naoqi_robot/jsk_nao_startup/launch/jsk_nao_startup.launch
@@ -60,5 +60,5 @@ enable_turbo_button: 0  # A button
 
   <node name="speaking_program_is_started_or_terminated"
         pkg="roseus" type="roseus"
-        args="$(find jsk_robot_startup)/lifelog/speaking-program-is-started-or-terminated.l &quot;(progn (unix:sleep 2) (good-morning) (do-until-key (ros::spin-once)))&quot;"/>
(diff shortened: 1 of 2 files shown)


pazeshun:
Just FYI:
> Before this change, it becomes over 80...

Maybe you can see one CPU load is 100% in this situation.
This is commonly known issue of looping without sleeping.
https://ja.wikipedia.org/wiki/%E3%83%93%E3%82%B8%E3%83%BC%E3%82%A6%E3%82%A7%E3%82%A4%E3%83%88
https://github.com/start-jsk/jsk_apc/pull/2236#issuecomment-313822965

kochigami:
@pazeshun 
Thank you for your information!

k-okada:
how about using `(ros::spin)` instead of `(do-until-key (ros::sleep)(ros::spin-once)` ?

kochigami:
@k-okada @pazeshun Thank you for your comment.

After using `ros:spin`,
> Note:
After the node is killed, NAO speaks good night, but Pepper doesn't.
I don't know why. I couldn't have solved the issue yet.

was solved. 
I could also learn the difference of `ros:spin` and `(do-until-key (ros::spin-once)...` a little.
`ros:spin`: will not return until node is shutdown => no need to sleep & easier to catch signal-handler
`ros:spin-once`: will return in every loop => need to sleep & not easier to catch signal handler

kochigami:
Travis passed.

PR #953, opened 2018-08-02
[jsk_robot_startup/base_trajectory_logger.py] Modified threshold of pos and theta.

iory:
This PR fixed the threshold of base_trajectory update.

Before this PR, calculating threshold become large value,
because ```dt``` is small.

https://github.com/jsk-ros-pkg/jsk_robot/blob/master/jsk_robot_common/jsk_robot_startup/src/jsk_robot_startup/lifelog/base_trajectory_logger.py#L129
https://github.com/jsk-ros-pkg/jsk_robot/blob/master/jsk_robot_common/jsk_robot_startup/src/jsk_robot_startup/lifelog/base_trajectory_logger.py#L131

Also, we should use threshold value that easy to understand.
Before this PR, the function ```diff_pose``` returns  the difference postion in mm order and absolute distance of quaternion (this value's unit is not intuitive) .
Therefore, I modified ```diff_pose``` to return angle[rad] of two poses, changed ```diff_pos``` in mm order, changed ```diff_theta``` in deg order, 
 and add param of ```thre``` and ```rthre```.

### DIFF
--- a/jsk_robot_common/jsk_robot_startup/src/jsk_robot_startup/lifelog/base_trajectory_logger.py
+++ b/jsk_robot_common/jsk_robot_startup/src/jsk_robot_startup/lifelog/base_trajectory_logger.py
@@ -126,7 +126,7 @@ class BaseTrajectoryLogger(LoggerBase):
                 self.latest_pose = self.get_pose_from_tf()
             if self.latest_pose:
                 if prev_pose:
-                    dt = (rospy.Time.now() - self.latest_pose.header.stamp).to_sec()
+                    dt = (rospy.Time.now() - prev_pose.header.stamp).to_sec()
                     if dt > 0:
                         thre = 0.1 + 1.0 / dt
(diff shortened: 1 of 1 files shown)


furushchev:
Thanks for the fix!! Basically LGTM.
I'll check with PR2 tomorrow.

And just a question: how did you choose default threshold values?

iory:
Please merge this.

k-okada:
http://www.ros.org/reps/rep-0103.html expecting SI units for ROS programs, and for example 
http://wiki.ros.org/amcl  uses `[m]` for most (?) all (?) parameters. Is there any problem if we use `[m]` instead of `[mm]`?

iory:
Considering your opinion, using SI units for this program.
Please check it.

k-okada:
thank you . BTW, how about add launch files for these scripts so that we can see the documentation with `--ros-args` or, update `roscore` to support `rosparm docs`, `rosparm help` or something

PR #973, opened 2018-10-12
jsk_robot_lifelog: add option to set custom collection for recording topics

furushchev:
CC: @Affonso-Gui


Affonso-Gui:
This allows organizing data from different experiments or from different topics into different collections, being easier to handle / query / convert. When converting mongodb json data into hd5 files, for instance, having the data into different collections will avoid having to convert unnecessary 600GB of data.

It will be useful for Anne's experiments and future lifelog usage.

@knorth55

k-okada:
@Affonso-Gui BTW, what is Anne’s github account ?

Affonso-Gui:
@k-okada Apparently she does not have any active account (?)

PR #988, opened 2018-11-08
[WIP] [jsk_pr2_robot] Add PR2 teleop command list

YutoUchimi:


furushchev:
I recommend to remap `/teleop/joy_vel` to `/base_controller/command_unchecked` for teleop node and just remove the `rosrun` part from the pics.

furushchev:
The comment above is wrong, sorry, please revert the change.
You might have to create a node to dynamically switch `/teleop/joy_vel` to `/base_controller/command_unchecked` only when follow hand mode instead.

knorth55:
@YutoUchimi what is status of this issue?

YutoUchimi:
I reverted the remap commit and pushed only document addition commit.
The remapping should be fixed in another PR, so this PR is ready to be merged.

PR #1008, opened 2018-11-29
[fetch] Make .rosinstall for both robot PC and user PC

708yamaguchi:
- In this Pull Request, I made `.rosinstall`  for both robot PC and user PC.
The former is `jsk_fetch.rosinstall` and the latter is `jsk_fetch_user.rosinstall.indigo`  and `jsk_fetch_user.rosintall.kinetic`.

 - Before this pull request, tar file of `fetch_ros` is installed to `kinetic` user PC by `wstool`, but tar file is inconvenient because it is not git repository.
After this pull request, git repository of `fetch_ros` will be installed to `kinetic` user PC by `wstool`.

 - In `jsk_fetch_user.rosinstall.kinetic`, `fetch_ros` version is `0.7.14` because my commit (fetchrobotics/fetch_ros@c0bc8bc) is merged and fetch's amcl will be improved.

### DIFF
--- a/jsk_fetch_robot/README.md
+++ b/jsk_fetch_robot/README.md
@@ -43,9 +43,9 @@ mkdir -p catkin_ws/src
 cd  catkin_ws
 wstool init src
 if [ $ROS_DISTRO = "indigo" ]; then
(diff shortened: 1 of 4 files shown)


knorth55:
fetch robot PC uses fetch-robotics apt server for package maintenance.
That's also another reason to have rosinstall for robot and users separately.

PR #1027, opened 2019-02-16
[peppereus] add default poses to pepper.yaml

kochigami:
I created some default poses for Pepper because it is hard to make some poses every time I create a demo :)


kochigami:
@k-okada 

k-okada:
Sorry for late. I saw the pictures of your presentation at MirIkan and tha is super nice!

PR #1031, opened 2019-02-27
[jsk_naoqi_robot/README] Add information mentioned at #1030

kochigami:

PR #1038, opened 2019-03-22
[jsk_nao_startup] add joy-client.l to jsk_nao_startup.launch

kochigami:
I want to include same nodes for nao and pepper, so I added `joy-client.l` to `jsk_nao_startup.launch`.

TODO
- [x] try this with real robot (nao and pepper): (memo) I don't know why two variables `*power-button-state*` and `power-button-state` are required. Let's check it!
- [x] consider #1046 
~- [ ] create button map of a controller, add it to README~ => #1031

### DIFF
--- a/jsk_naoqi_robot/jsk_nao_startup/nodes/joy-client.l
+++ b/jsk_naoqi_robot/jsk_nao_startup/nodes/joy-client.l
@@ -1,0 +1,88 @@
+#!/usr/bin/env roseus
+
+
+(ros::roseus "joy_client")
+(ros::load-ros-package "naoeus")
+
+
+(if (not (find-package "NAOQI_BRIDGE_MSGS"))
+    (make-package "NAOQI_BRIDGE_MSGS" :nicknames "NAOQI_MSGS"))
+
+(ros::advertise "/speech" std_msgs::String 1)
+(defun speak-jp (str &rest args)
+  (let ((msg (instance std_msgs::String :init)))
+    (send msg :data (apply #'format nil str args))
+    (ros::ros-info "speech ~A" (send msg :data))
+    (ros::publish "/speech" msg))
+  )
+
+(ros::advertise "/nao_tweet" std_msgs::String 1)
+(defun tweet (str &rest args)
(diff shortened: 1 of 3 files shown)


jsk_naoqi_robot/jsk_nao_startup/nodes/joy-client.l
```diff
@@ -0,0 +1,88 @@
...
+			   (axes (send msg :axes))
+			   (power-button-state (elt buttons 8))
+			   (rb-button-state (elt buttons 5))
+			   (b-button-state (elt buttons 1))
+			   (x-button-state (elt buttons 2))
+			   (y-button-state (elt buttons 3)))
+		      (print (list buttons power-button-state (ros::time- (ros::time-now) *power-button-time*)))
+		      (cond ((and (= *power-button-state* 1)
+				  (=  power-button-state  0)
+				  (> (send (ros::time- (ros::time-now) *power-button-time*) :to-sec) 2))
```
kochigami: This condition will never be satisfied because `*power-button-time*` is updated in each loop.

jsk_naoqi_robot/jsk_nao_startup/nodes/joy-client.l
```diff
@@ -0,0 +1,88 @@
...
+		#'(lambda (msg)
+		    (let* ((buttons (send msg :buttons))
+			   (axes (send msg :axes))
+			   (power-button-state (elt buttons 8))
+			   (rb-button-state (elt buttons 5))
+			   (b-button-state (elt buttons 1))
+			   (x-button-state (elt buttons 2))
+			   (y-button-state (elt buttons 3)))
+		      (print (list buttons power-button-state (ros::time- (ros::time-now) *power-button-time*)))
+		      (cond ((and (= *power-button-state* 1)
```
kochigami: Both `*power-button-state*` and `power-button-state` are required because we need to check the change of button value.
If only `power button status` is checked, servo on / off will be performed many times.

kochigami:
@k-okada 

PR #1015, opened 2018-12-04
[WIP] [naoqi] allow group namespace to naoqi/pepper/nao-interface.l 

kochigami:
new version of #922  

I also added doc and sample.
I need to try with real robot.

PR #1039, opened 2019-03-24
[naoqieus] add group-namespace to speak-action method

kochigami:
This change enables us to use `:speak-action` method with multiple naoqi robots separated by group-namespace (ref. #1015).


kochigami:
@k-okada 

PR #1042, opened 2019-03-28
add rqt_pr2_dashboard as dependency

knorth55:


knorth55:
pr2_rqt_dashboard is required to run PR2Dashboard desktop app.

PR #1048, opened 2019-04-17
[jsk_robot_startup] Add option to check isAlive for lightweight_logger

708yamaguchi:
In this Pull Request, I added an option to ignore `vital_checker_->isAlive`.

Current problem is below:

`joint_states_logger` output error when `joint_states_throttle/output` is not subscribed for long time.
https://github.com/jsk-ros-pkg/jsk_robot/blob/master/jsk_robot_common/jsk_robot_startup/lifelog/common_logger.launch#L195-L204

However, `joint_states_throttle/output` is published only when robots move.

Therefore, robots using `joint_states_logger` (fetch or PR2) output error whenever they are not moving.

I think this PR solve this problem.

### DIFF
--- a/jsk_fetch_robot/jsk_fetch_startup/launch/fetch_bringup.launch
+++ b/jsk_fetch_robot/jsk_fetch_startup/launch/fetch_bringup.launch
@@ -32,6 +32,7 @@
   <!-- logging -->
   <include file="$(find jsk_fetch_startup)/launch/fetch_lifelog.xml">
     <arg name="map_frame" value="$(arg map_frame)" />
+    <arg name="vital_check" value="false" />
   </include>
 
   <!-- diagnostic aggregator -->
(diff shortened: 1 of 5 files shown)


knorth55:
This problem happens in PR2, too.
`joint_states_throttle` in `common_logger.launch` publishes only when `joint_states` have difference from the last posture with param `periodic: false`.
This means that `joint_state_throttle` publish only when robot is moving.
https://github.com/jsk-ros-pkg/jsk_robot/blob/master/jsk_robot_common/jsk_robot_startup/lifelog/common_logger.launch#L186-L194
https://github.com/jsk-ros-pkg/jsk_robot/blob/master/jsk_pr2_robot/jsk_pr2_startup/jsk_pr2_lifelog/db_client.launch#L28-L34

`vital_checker_->isAlive` is checked if new topic is coming in every 1 second.
Therefore, `joint_states_throttle` with `periodic: false` does publish the topic when robot is not moving and `vital_checker_->isAlive` returns false so that `LightweightLogger` shows error below almost every 1 second.
With this PR, we can suppress this error.

diagnostic

PR #1050, opened 2019-04-17
[Fetch] Remove /tuck_arm which is already launched in fetch_bringup/launch/include/teleop.launch.xml

708yamaguchi:
Now in Fetch, `/tuck_arm` is launched in both
- fetch_teleop.xml in `jsk_fetch_startup` (https://github.com/jsk-ros-pkg/jsk_robot/blob/master/jsk_fetch_robot/jsk_fetch_startup/launch/fetch_teleop.xml#L28 )

and

- teleop.launch.xml in `fetch_bringup` (https://github.com/fetchrobotics/fetch_robots/blob/indigo-devel/fetch_bringup/launch/include/teleop.launch.xml#L18 )

In this PR, we remove `/tuck_arm` in `jsk_fetch_startup`.

### DIFF
--- a/jsk_fetch_robot/jsk_fetch_startup/launch/fetch_teleop.xml
+++ b/jsk_fetch_robot/jsk_fetch_startup/launch/fetch_teleop.xml
@@ -25,8 +25,6 @@
 
   <node name="controller_reset" pkg="fetch_bringup" type="controller_reset.py" />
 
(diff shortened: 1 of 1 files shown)


k-okada:
why we need to run two `tuck_arm`?

708yamaguchi:
We do not want to run two /tuck_arm, so I remove one of two /tuck_arm.

2019年4月18日(木) 12:08 Kei Okada <notifications@github.com>:

k-okada:
i see

PR #1062, opened 2019-04-19
[fetch] Enable arm teleop and set buttons similar to PR2

708yamaguchi:
Before, we disable fetch's arm teleop, but this causes error (sometimes fetch move base and sometimes move arm).
To prevent this error, we assign arm teleop buttons to unused buttons.

Also, I change button assignment for head control similar to PR2.

PR #1063, opened 2019-04-19
[fetch] update jsk_fetch_startup/README.md to add admin modification info

knorth55:
We disable teleop nodes in `/etc/ros/indigo/robot.launch` because `jsk_fetch_startup/fetch_bringup.launch` have its own teleop nodes.

cc. @708yamaguchi

### DIFF
--- a/jsk_fetch_robot/jsk_fetch_startup/README.md
+++ b/jsk_fetch_robot/jsk_fetch_startup/README.md
@@ -117,6 +117,10 @@ Logging script is initialized at:
 - 2019/04/19: set `env DISALLOW_MODULE_LOADING=0` in `/etc/init/pulseaudio.conf`.
   - this modification is needed for overriding default speaker setting in `/etc/init/jsk-fetch-startup.conf`
   - overriding default speaker setting to use USB speaker on head is done with `pactl set-default-sink $AUDIO_DEVICE` in `/etc/init/jsk-fetch-startup.conf`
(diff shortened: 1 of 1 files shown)


708yamaguchi:
thank you for updating readme

knorth55:
@708yamaguchi thank you, i updated.

knorth55:
@k-okada 
We need to set `launch_teleop:=false` for `fetch_bringup/fetch.launch` because we remap topic for unsafe teleop.
That's why we made our own `teleop.launch` to support unsafe teleop (by @furushchev )
https://github.com/jsk-ros-pkg/jsk_robot/blob/master/jsk_fetch_robot/jsk_fetch_startup/launch/fetch_teleop.xml#L13

Anyway i will add comment about why we need to use our own teleop launch.

knorth55:
@k-okada I update this PR #1062 .
In the end, we need to replace most teleop nodes to run our unsafe and safe teleop.
That is why we need to set `launch_teleop:=false` for `fetch_bringup/launch/fetch.launch`
https://github.com/jsk-ros-pkg/jsk_robot/blob/e35bb4b8d234f6a0083f2fb2eb1a07acb60fc7d5/jsk_fetch_robot/jsk_fetch_startup/launch/fetch_teleop.xml

knorth55:
@k-okada can you review and merge this PR?

Issue #1060, opened 2019-04-18
[pr1040] respeaker node does not start


knorth55:
USB connector is not good.
we need to redesign the attachment

k-okada:
where is the cad data?
we should put them in
https://github.com/jsk-ros-pkg/jsk_robot/tree/master/jsk_pr2_robot/jsk_pr2_accessories
--
◉ Kei Okada



2019年4月19日(金) 3:16 Shingo Kitagawa <notifications@github.com>:

knorth55:
I add more information about this issue.
this issue happens because USB connecter in ReSpeaker mic got loose.
Current attachment CAD is in Google team drive.
Should I manage it in github?

k-okada:
> Should I manage it in github?

yes
--
◉ Kei Okada

2019年4月19日(金) 12:22 Shingo Kitagawa <notifications@github.com>:

knorth55:
OK i will upload all CAD data.

knorth55:
I upload all cad data https://github.com/jsk-ros-pkg/jsk_robot/pull/1064

knorth55:
if this issue happens, please open this issues again.

PR #1069, opened 2019-04-24
[pr2] use new respeaker mount

knorth55:
I redesign PR2 head respeaker mount and made it with 3D printer (dimension).


knorth55:
This mount may solve #1060 

Previous mount:
IMG_4181

New mount:
IMG_4180

New mount with PR2
IMG_4179

knorth55:
why i remake this mount is that previous mout looked really bad.
i put longer flat space near cable commection so that we can tie the cable with band and prevent disconnection.

PR #1068, opened 2019-04-22
[baxtereus] Pass start-time to start-offset-time of :angle-vector-motion-plan

pazeshun:
Updated version of #956 
Includes #1067 

This PR fixes the issue detected by #1067 with the same method as `fetcheus`
cf. https://github.com/jsk-ros-pkg/jsk_robot/blob/master/jsk_fetch_robot/fetcheus/fetch-interface.l#L79
https://github.com/jsk-ros-pkg/jsk_robot/blob/master/jsk_fetch_robot/fetcheus/fetch-interface.l#L82
https://github.com/jsk-ros-pkg/jsk_robot/blob/master/jsk_fetch_robot/fetcheus/fetch-interface.l#L91

PR #1070, opened 2019-04-24
[fetcheus] support :start-offset-time in fetcheus angle-vector and angle-vector-sequence

knorth55:
Fetch version of #1068 
Fetch support both `:start-time` and `start-offset-time` API as `baxtereus` in #1068.

- change not to ignore `:start-offset-time` 
- test `start-time` and `start-offset-time`


knorth55:
I tested with real robot, and it works.

knorth55:
Gazebo simulation is something weird... :thinking: 
When the robot wait for motion start, motors of the robot probably got servo off.

- :100: : `(send *ri* :angle-vector (send *fetch* :angle-vector) 2000 nil 0)`
fetch-moveit-start-time-0
- :thinking: :`(send *ri* :angle-vector (send *fetch* :angle-vector) 2000 nil 4)`
fetch-moveit-start-time-4

pazeshun:
Question: Will servo off cause travis test failure?
Suggestion: Could you confirm new test fails in current code in another PR?

knorth55:
> Question: Will servo off cause travis test failure?
Suggestion: Could you confirm new test fails in current code in another PR?

No, and I will fix this problem.

pazeshun:
> Suggestion: Could you confirm new test fails in current code in another PR?

I made #1071

knorth55:
I made PR to fix this fetch servo off problem.
https://github.com/fetchrobotics/robot_controllers/pull/40

PR #1080, opened 2019-05-21
[jsk_pepper_startup] tweet when jsk_pepper_startup nodes are created (updated version of #1074)

kochigami:
- delete naoqi_dashboard and naoqi_apps (kochigami-develop) because I'd like to avoid creating an extra test launch
- enable to add Pepper's name when tweeting "good morning"

~まだ実機で試していません~


kochigami:
@k-okada 

PR #1081, opened 2019-05-22
[WIP] PR1040 Develop branch

knorth55:
This branch is used in `PR1040` `applications` user.
Current `PR1040` uses this branch.
split into #1082, #1083,  #1084, #1084, #1085, #1085, #1086, #1087, #1088, #1089, #1090, #1091, #1092 , #1121 

- big changes
  - [x] merge tweet change into `origin/master`
  - [x] battery change into `origin/master` (#1082)
  - [x] `hark` -> `respeaker` (#1085)
  - [ ] navigation change into `origin/master`
  - [ ] logger changes (mainly change in `LightweightLogger` #1121 )

cc. @furushchev , @YutoUchimi

PR #1082, opened 2019-05-22
[jsk_pr2_startup] use new battery warning

knorth55:
split from #1081 
this battery warning is written by @furushchev and used in PR1040.
cc. @YutoUchimi


knorth55:
@YutoUchimi 
in this document, we need at least 56V.
so it is better to output error when voltage is lower than 14V ( 56 / 4).
http://www.clearpathrobotics.com/wp-content/downloads/pr2//pr2_power_board_rr.pdf

knorth55:
@YutoUchimi 
also, you can get voltage info from `pr2_power_board`.
when PR2 is down because of power shortage, voltage of CPU, arms and base is around 59-60V.
http://wiki.ros.org/pr2_power_board

YutoUchimi:
Note: battery log (2019.5.23) is here.
https://drive.google.com/drive/u/1/folders/14QrFi53zJJgZTEjqmford_730Xaii3Ra

k-okada:
@YutoUchimi Is this depends on (or related to) https://github.com/jsk-ros-pkg/jsk_common/pull/1628 ? Or these two tools are completely different system?

YutoUchimi:
@k-okada They are completely different.

PR #1088, opened 2019-05-22
launch all sound play nodes in pr2_bringup.launch

knorth55:
split from #1081 
launch all sound play nodes in `pr2_bringup.launch`
cc. @furushchev


k-okada:
do we really need this? My understanding is; this change enable PR2 robot to speak Japanese before we launch `jsk_pr2_startup pr2.launch`. If this is correct, I do not think we need this, if there is some node that start before `jsk_pr2_startup pr2.launch` and speaks in Japanese, then that node should speak in English.

knorth55:
@k-okada 
Probably, you are right.
If PR1040 works fine without this PR, i will close this PR.

PR #1090, opened 2019-05-22
[jsk_pr2_startup] fix ntp monitor in pr2_bringup.launch

knorth55:
split from #1081 
disable ntp monitoer in `pr2_bringup.launch`
cc. @furushchev


k-okada:
why did you remove this. I think time synchronization between c2 and c2 is veery important...

knorth55:
i have no idea why it is disabled in `PR1040` but it is disabled by @furushchev .
i will test if it works with `launch_ntp_monitor:=true`.

knorth55:
i update to fix ntp monitor, and now it works!
ntp_robot_monitor

knorth55:

PR #1094, opened 2019-05-22
[jsk_robot_startup] load-ros-manifest pr2_mechanism_msg and fetch_driver_msgs only when they exist in active_user.l

knorth55:
split from #1081 
add missing `run_depend` `pr2_mechanizm_controllers` in `jsk_robot_startup`
cc. @furushchev


k-okada:
which nodes in jsk_ronot_startup depends on pr2 mechanisms

knorth55:
@k-okada 
`jsk_robot_startup/lifelog/active_user.l` depends on `pr2_mechanism_msgs`
https://github.com/jsk-ros-pkg/jsk_robot/blob/master/jsk_robot_common/jsk_robot_startup/lifelog/active_user.l#L4

k-okada:
ok, in this case, please remove `pr2_mechanism_controllers` from `run_depend` and change code like
```
(unless (ros::rospack-find "pr2_controllers_controllers")
  (ros::load-ros-manifest "pr2_mechanism_controllers"))
```
so that is users install `jsk_robot_startup` directory, they can use without `pr2_mechanism_controllers` and if users install `jsk_robot_startup` from `jsk_pr2_startup`, then then can use with `pr2_mechanism_contreollers`
Because, we do not want to make dependency from `common` to `pr2`.

knorth55:
@k-okada you mean `load-manifest` only when the package exists like this?
```
(when (ros::rospack-find "pr2_mechanism_msgs")
  (ros::load-manifest "pr2_mechanism_msgs"))
```

k-okada:
yes
--
◉ Kei Okada



2019年5月23日(木) 20:40 Shingo Kitagawa <notifications@github.com>:

knorth55:
I update to load `pr2_mechanism_msgs` `fetch_driver_msgs` when the package is already installed.

PR #1097, opened 2019-05-27
Add teleop with euslisp

itohdak:
I added a euslisp file for teleoperation with ps3joy.
Since I didn't know where to put this file, I temporarily added the file under jsk_robot_utils.
Also this file is written only for PR2, so it should be generalized so that we can use this with other robots too.
@knorth55 

This file is originally written for HRP2W, and it looks like [this](https://gist.github.com/itohdak/53e768a2a542a86e1e6e579dc8e47a7a).


itohdak:
I renamed `pr2-teleop.l` to `robot-teleop.l`, and removed pr2 depending codes.
I also added some examples with pr2 robot and hrp2jsknts robot using `robot-teleop.l`.
I tended to follow the commands in [jsk_pr2_robot](https://github.com/jsk-ros-pkg/jsk_robot/tree/master/jsk_pr2_robot#jsk_pr2_robot) as much as possible.
To avoid the commands from colliding with pr2 default controller commands, I provided some options to toggle on/off of some callbacks.

knorth55:
this is for teleoperation with joystick, right?
in that case, can you rename the file to `robot-teleop-joy.l`?

itohdak:
@knorth55 
Thank you, that's right.
I renamed the filenames and the variables from `teleop` to `teleop-joy`.

PR #1101, opened 2019-06-19
add dock/undock app for fetch

knorth55:
add dock/undock app for fetch.
this app is same as controller dock/undock button.
Now we can run this app on Fetch15 with voice command.

cc. @708yamaguchi


knorth55:
@Affonso-Gui this is Fetch dock/undocking simplest app.

knorth55:
demo video: https://drive.google.com/open?id=1UPGm5sL_aFWPgf0DFoGKb53FLrpH-zfl

knorth55:
I updated.
Dock

Undock

PR #1113, opened 2019-07-12
add inverse-kinematics-raw in baxter-robot

knorth55:
add original `:inverse-kinematics-raw` method in `baxter-robot`


knorth55:
@k-okada i want to use this for openlab, so please merge this

PR #1116, opened 2019-07-23
Make rviz node to anonymous

iory:
In current, Fetch and PR2's ```rviz.launch``` node is not anonymous.
If mutliuser launch this launch file to view robot state, only one node can be launched.
So I modified launch file to launch anonymous node.


708yamaguchi:
result of `roslaunch rviz.launch` is below:
```
$ rosnode list | grep rviz
/jsk_startup_rviz_b4apc_13616_1257488453476967856
/rviz_1563545127204263987
/rviz_1563765973520799457
/rviz_1563769778509227934
/rviz_1563770784468091842
```

k-okada:
another method to check the node name (without launching)
```
$ roslaunch --nodes jsk_pr2_robot/jsk_pr2_startup/rviz.launch 
/jsk_startup_rviz_p51s_28868_5504708204869271610
```

PR #1043, opened 2019-03-29
[WIP] Furushchev Develop

knorth55:
FYI, furushchev development branch is this.
@Affonso-Gui @iory

PR #1121, opened 2019-08-17
fix to work LightweightLogger

knorth55:
Split from #1043 
This PR needs to work `LightWeightLogger` properly.
Except this change, `LightWeightLogger` log nothing.

cc. @furushchev


k-okada:
this fails on `ROS_DISTRO=indigo USE_DEB=false`, 
```
Errors << jsk_robot_startup:make /home/travis/ros/ws_jsk_robot/logs/jsk_robot_startup/build.make.000.log
/home/travis/ros/ws_jsk_robot/src/jsk_robot/jsk_robot_common/jsk_robot_startup/src/lightweight_logger_nodelet.cpp: In member function 'virtual void jsk_robot_startup::lifelog::LightweightLogger::inputCallback(const ros::MessageEvent<const topic_tools::ShapeShifter>&)':
/home/travis/ros/ws_jsk_robot/src/jsk_robot/jsk_robot_common/jsk_robot_startup/src/lightweight_logger_nodelet.cpp:189:44: error: 'class mongodb_store::MessageStoreProxy' has no member named 'getNumInsertSubscribers'
       if (!wait_for_insert_ && msg_store_->getNumInsertSubscribers() == 0) {
                                            ^
make[2]: *** [CMakeFiles/jsk_robot_lifelog.dir/src/lightweight_logger_nodelet.cpp.o] Error 1
make[1]: *** [CMakeFiles/jsk_robot_lifelog.dir/all] Error 2
make: *** [all] Error 2
cd /home/travis/ros/ws_jsk_robot/build/
```

knorth55:
we need this feature for `mongodb_store` `0.4.3` at least.
this version is only release in `kinetic`. (`melodic` is also on its way https://github.com/strands-project/mongodb_store/pull/244).

jsk_robot_common/jsk_robot_startup/CMakeLists.txt
```diff
@@ -29,7 +29,11 @@ find_package(catkin REQUIRED COMPONENTS
   urdf
 )
 
-find_package(Boost REQUIRED)
+find_package(Boost REQUIRED COMPONENTS program_options thread system)
+if ("${BOOST_MAJOR_VERSION}" VERSION_EQUAL "1" AND "${Boost_MINOR_VERSION}" VERSION_GREATER "46")
```
YutoUchimi: When major version of boost is upgraded to 2 or upper, this won't work correctly.

knorth55: Who uses boost upper 2.0?
REP 3 does not support Boost 2.0 https://www.ros.org/reps/rep-0003.html#melodic-morenia-may-2018-may-2023

YutoUchimi: I know, but it doesn't mean jsk_robot won't support boost>=2.0 __forever__.
Also, that's not the point.
You dislike this long boolean comparison line, right?
```
if ("${Boost_MAJOR_VERSION}.${Boost_MINOR_VERSION}.${Boost_SUBMINOR_VERSION}" VERSION_EQUAL "1.47.0" OR
    "${Boost_MAJOR_VERSION}.${Boost_MINOR_VERSION}.${Boost_SUBMINOR_VERSION}" VERSION_GREATER "1.47.0")
```
Actually I would like to use `VERSION_GREATER_EQUAL`, but it can only be available in cmake>=3.7
So I wrote a logical alternative like this.

knorth55: I don't get why you need to concatenate all version numbers into one.
You can compare these numbers separately.
Also, We haven't tested on Boost>=2.0,  so `BOOST_MAJOR_VERSION` should be `1`.

YutoUchimi: OK, I removed `Revert "use chrono in boost >= 1.47.0"` commit.

PR #1132, opened 2019-09-16
[aero][wip] add jsk_aero_teleop

taichiH:
I added aero_teleop.
That enables us to switch default teleop (aero-ros-pkg) and original ik teleop.

I can apply this program https://github.com/jsk-ros-pkg/jsk_robot/pull/1097 to aero_ik_teleop because both are almost same.
I will change `aero-teleop.l` to `jsk_robot_common/jsk_robot_utils/euslisp/aero-teleop-joy.l ` after https://github.com/jsk-ros-pkg/jsk_robot/pull/1097 merge.


k-okada:
merged @itohdak 's euslisp teloep code in #1197
In this case, you can add your code on top of #1197 and send PR

PR #1133, opened 2019-09-16
add aero_speech nodes

taichiH:
I added speech_recognition nodes.
That enables aero to speak and listen to Japanese.


taichiH:
We always expect a robot to have some intelligent response/reaction when talking to the robot.

As an example of reaction, physical movements are also considered.
But it is dangerous and undesired to move the robot when unexpected timing because of misrecognition of speech. so I thought that speaking weather might be good.

But I also think that it is better to delete the function from the viewpoint of maintenance.
If you think that it's better for remove it, I remove it.

taichiH:
There is no need to speak the weather to check if speech_recognition is working properly, so I removed the `speak_weather_info` function.

PR #1137, opened 2019-10-02
[jsk_fetch_startup] add battery warning node to fetch

HiroIshida:
Robot warns if the battery percentage is under some threshold.
@knorth55


HiroIshida:
under testing now in the real robot fetch 15

PR #1140, opened 2019-10-03
[jsk_fetch_startup] respawn fetch sound_play for audio_common bug

knorth55:
we need to set sound_play respawn=true for audio_common bug.
cc. @708yamaguchi @itohdak

### DIFF
--- a/jsk_fetch_robot/jsk_fetch_startup/launch/fetch_bringup.launch
+++ b/jsk_fetch_robot/jsk_fetch_startup/launch/fetch_bringup.launch
@@ -24,6 +24,10 @@
   </node>
   <node if="$(arg boot_sound)" pkg="jsk_fetch_startup" name="boot_sound" type="boot_sound.py" />
 
+  <!-- english speach node -->
+  <node name="sound_play" pkg="sound_play" type="soundplay_node.py"
+        respawn="true" if="$(arg launch_sound_play)" />
+
   <!-- japanese speech node -->
   <include file="$(find voice_text)/launch/voice_text.launch">
     <arg name="launch_sound_play" value="$(arg launch_sound_play)" />
@@ -85,7 +89,7 @@
         args="0 0 0.1 0 0 0 head_pan_link respeaker_base 100"/>
   <include file="$(find julius_ros)/launch/julius.launch">
     <arg name="launch_audio_capture" value="false"/>
-    <arg name="launch_sound_play" value="$(arg launch_sound_play)"/>
+    <arg name="launch_sound_play" value="false"/>
     <arg name="speech_to_text_topic" value="speech_to_text_julius"/>
   </include>
   <include file="$(find respeaker_ros)/launch/sample_respeaker.launch">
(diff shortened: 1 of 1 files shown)


knorth55:
we use https://github.com/itohdak/audio_common/tree/0.3.3%2Bbugfix this branch to avoid stopping `audio_common`.

knorth55:
@HiroIshida @708yamaguchi Is Fetch15 speaking these days? I think he is speaking correctly these days.

708yamaguchi:
I heard fetch's 「ツイートしました」 yesterday.

I think there is no problem.

itohdak:
FYI, I made a PR (https://github.com/ros-drivers/audio_common/pull/126) to `audio_common` in order to fix the robot's stop-speaking bug.
According to my investigation, I found that the GStreamer objects, which are created to produce sounds, were not correctly released during the deletion and were causing some memory leaks.

I would like you to try this PR and check if it works when you have time.

HiroIshida:
I think he speaks correctly, I heard he says something when battery changes from 90 to 89 and 80 to 79 these days

jsk_fetch_robot/jsk_fetch_startup/launch/fetch_bringup.launch
```diff
@@ -77,12 +80,16 @@
...
+  <node name="sound_play" pkg="sound_play" type="soundplay_node.py"
+        respawn="true" if="$(arg launch_sound_play)" />
+
   <!-- speech recognition -->
   <node name="respeaker_transformer" pkg="tf" type="static_transform_publisher"
         args="0 0 0.1 0 0 0 head_pan_link respeaker_base 100"/>
   <include file="$(find julius_ros)/launch/julius.launch">
     <arg name="launch_audio_capture" value="false"/>
-    <arg name="launch_sound_play" value="$(arg launch_sound_play)"/>
+    <arg name="launch_sound_play" value="false"/>
```
k-okada: why you set "false" here? instead of `$(arg launch_sound_play)`

knorth55: we change to launch `sound_play` outside of `julius` launch.
i think this is easier to understand.
https://github.com/jsk-ros-pkg/jsk_robot/pull/1140/files/297f87073395eb72257fde00e0f3fb24c65622e5#diff-f9ac819e72319dde53123b0b9a7ffca8R84-R86

k-okada:
ok, please add that comment on launch file
--
◉ Kei Okada



2019年10月11日(金) 15:06 Shingo Kitagawa <notifications@github.com>:

knorth55:
I updated and add comments.

708yamaguchi:
FYI

If the fetch's USB speaker is unplugged, the USB speaker will not be reconnected (fetch is trying to use unavailable speaker). We checked this in System Settings -> Sound in Ubuntu.
This problem also occurs when the fetch gets over the step.
We maybe need to fix the speaker to be used.

fetchのUSBスピーカーが抜き差しされると、fetchは使用できないスピーカーを使用しようとします。UbuntuのSystem Settings -> Soundから確認しました。
この問題はfetchが段差を乗り越えるときにも発生している気がして、USBスピーカーが一瞬抜けて次の瞬間に再接続する際に、fetchが喋らなくなっている気がします。
fetchが使用するスピーカーを固定する必要があるかもしれません。

checked with @knorth55

PR #1151, opened 2019-10-18
[jsk_fetch_startup] add time_signal.py

knorth55:
I don't know who, but someone left this script in fetch15 home directory.
so I move this script into jsk_robot.
this node is added in `crontab` as below.
```
0 * * * * /home/fetch/ros/indigo/devel/env.sh rosrun jsk_fetch_startup time_signal.py
```

cc. @708yamaguchi @iory


708yamaguchi:
I am sorry, but I left this script to fetch's home directory. In my memory, this script works correctly.

However, some people say, "I feel like being rushed when fetch speaks time signal."
Then, I have no idea whether this script should be used...

knorth55:
lets save this script in jsk_robot first.
then think about how to use this later.

knorth55:
we run this script for a week and no one claim about it.

PR #1155, opened 2019-10-22
[fetcheus] Add test of :go-velocity

pazeshun:
Add test to catch  #1154

PR #1156, opened 2019-10-22
[fetcheus] Fix :go-velocity to reject non-zero y correctly

### LINKED ISSUE #1154: [fetcheus] go-velocity does not work
```
(send *ri* :go-velocity 0.1 0 0.5)
[ERROR] [1571647540.324682228]: Setting velocity in y-axis is not valid for non-holonomic robot
nil
```

cc @Affonso-Gui  @YutoUchimi

pazeshun:
Includes #1155 
Fixes #1154

### DIFF
--- a/jsk_fetch_robot/fetcheus/fetch-interface.l
+++ b/jsk_fetch_robot/fetcheus/fetch-interface.l
@@ -209,7 +209,7 @@ Example: (send self :gripper :position) => 0.00"
     (x y d ;; [m/sec] [m/sec] [rad/sec]
        &optional (msec 1000) ;; msec is total animation time [msec]
        &key (stop t) (wait t))
-    (unless (not (zerop y))
+    (unless (zerop y)
       (ros::ros-error
         "Setting velocity in y-axis is not valid for non-holonomic robot")
       (return-from :go-velocity nil))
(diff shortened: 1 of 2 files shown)


pazeshun:
Travis test correctly failed in #1155, and correctly succeeded in this PR.
@knorth55 Could you review this?

pazeshun:
> This PR is not checked on real robot and causes serious error in https://github.com/jsk-ros-pkg/jsk_pr2eus/blob/master/pr2eus/robot-interface.l#L1433 !!!
> Now fetch is always failed to move.

What is the error?

knorth55:
`:go-velocity` returns `nil` everytime, so `:move-to` always fail.
so robot fails going to the kitchen.

knorth55:
I'm fixing now.

pazeshun:
> `:go-velocity` returns `nil` everytime, so `:move-to` always fail.
> so robot fails going to the kitchen.

I see.
So `:move-to-wait` should be updated for non-holonomic robots, or `:go-velocity` should accept non-zero `y` and just ignore it.
Sorry for your trouble.

knorth55:
yes.
I'm wondering which is better.

pazeshun:
How about printing warn message when non-zero y is passed?

pazeshun:
This solution may not good if you want to make strict navigation because assuming non-zero y is wrong...

knorth55:
I opened PR to fix the error https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/405

PR #1152, opened 2019-10-19
[WIP] Merge Fetch15 branch to master

knorth55:
Current fetch15 ws.
This branch have these features follows;

## Go to kitchen
- [x] fix fetch robot localization #1147 
- [ ] fix go-to-kitchen demo #1149 
- [x] add min abs velocity for :go-pos-unsafe #1159
- [ ] update auto-dock in navigation-utils.l #1160
- [ ] fix radian in `:go-pos-unsafe` #1166 
- [ ] set :correction nil in :move-to-wait for fetch #1169 

## Others
- [x] do not speak warning diagnostics to avoid noisy fetch #1128
- [ ] add fetch_ros in jsk_fetch.rosinstall #1148 
- [x] add time_signal.py #1151 
- [ ] do not set :wait-until-update when :worldcoords #1167
- [ ] fix ros-info in :go-pos-unsafe in fetch-interface.l #1168 
- [ ] use plugin for app_manager #1193 
- [x] add time_signal app #1194 
- [ ] fixed the odom topic of move_base #1206
- [ ] add base_camera_mount_mk1 models #1199 
- [ ] make jsk_accessories as ros package #1200
- [ ] fix install_upstart.sh #1204
- [ ] add jsk_fetch.xacro #1205
- [x] add `/head_camera/rgb/throttled/image_rect_color/compressed` #1219 
- [ ] use `pr2eus/speak.l` in `unsafe_warning.l` #1224 
- [ ] add rwt_image_view #1225
- [ ] add basic authentication #1232 
- [ ] common safe teleop system for JSK mobile robots #1234 

## Upstream depencency
- `jsk-ros-pkg/jsk_pr2eus`
  - [ ] fix :go-velocity problem (https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/406)
  - [x] add correction arg in :move-to-wait (https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/418)
- `fetchrobotics/fetch_ros`
  - [x] add topic args to navigation launch files (https://github.com/fetchrobotics/fetch_ros/pull/137)
  - [ ] add use_map_topic arg in fetch_nav.launch (https://github.com/fetchrobotics/fetch_ros/pull/144)
  - [ ] add launch_map_server argument in fetch_nav.launch (https://github.com/fetchrobotics/fetch_ros/pull/146)
- `tork-a/roswww`
  - [ ] add basic authentication in roswww (https://github.com/tork-a/roswww/pull/51)

PR #1160, opened 2019-10-25
[jsk_fetch_startup] update navigation-utils.l to work auto-dock correctly.

knorth55:
update `auto-dock.l` to work correctly.


k-okada:
LGTM, but why the fetch can autodock without this modification ? @708yamaguchi

708yamaguchi:
In `auto-dock.l`, \*ri\* is created when `navigation-utils.l` is loaded.
Therefore, `(require :fetch-interface "package://fetcheus/fetch-interface.l")` and `(fetch-init)` in this PR are not necessarily required.
https://github.com/jsk-ros-pkg/jsk_robot/blob/c3a53e1d8bf17585932316d1751d8bd949ba3869/jsk_fetch_robot/jsk_fetch_startup/euslisp/navigation-utils.l#L7-L10

However, I think this PR helps people to understand the code.

knorth55:
I mean adding `:clear-costmap` at the beginning is important to start this demo smoothly.

k-okada:
> In auto-dock.l, *ri* is created when navigation-utils.l is loaded.

ok, if so we need decide policy, whether if we should create `*ri*` within `navigation-utils.l` or not. If we decided to initialize `*ri*` outside `navigation-utils.l`, remove `(fetch-init)` within `navigation-utils.l` with comments and add `(fetch-init)` for all app programs loading `navigation-utils.l`.

knorth55:
i meant `:clear-costmap` is the most important point in this PR.
in order to do it, we first make *ri*.

knorth55:
I change to do `:clear-costmap` in `(auto-dock)` function in `navigation-utils.l`.

knorth55:
I'm using this function in #1152

knorth55:
@708yamaguchi can you review this again?

708yamaguchi:
@k-okada says
> ok, if so we need decide policy, whether if we should create *ri* within navigation-utils.l or not. If we decided to initialize *ri* outside navigation-utils.l, remove (fetch-init) within navigation-utils.l with comments and add (fetch-init) for all app programs loading navigation-utils.l.

I think making \*ri* inside of navigation-utils.l is no problem because we want to use many methods of \*ri* in navigation-utils.l.

Regarding `:clear-costmap`, if you think `:clear-costmap` is important,  how about using `:clear-costmap` in (go-to-spot) instead of (auto-dock)?

708yamaguchi:
thank you!
Looks good to me.

jsk_fetch_robot/jsk_fetch_startup/euslisp/navigation-utils.l
```diff
@@ -9,6 +9,21 @@
 (defparameter *spots* nil)
 (defparameter *dock-spot* "/eng2/7f/room73B2-fetch-dock-front")
 (defparameter *is-charging* nil)
+(defparameter *tfl* (instance ros::transform-listener :init))
```
k-okada: please add your `*ri*` policy here....

knorth55: I have no policy about `*ri*`.
This PR does not have  any policy about `*ri*`.
What I want to do is to do `:clear-costmap` before `:move-to`.

k-okada: https://github.com/jsk-ros-pkg/jsk_robot/pull/1160#issuecomment-546595519 changed the code based on policy, please add that to source code, also update https://github.com/jsk-ros-pkg/jsk_robot/pull/1160#issue-332598336. 
we would like to why this commit makes auto dock work correctly, too.

knorth55: I mean https://github.com/jsk-ros-pkg/jsk_robot/pull/1160#issuecomment-546595519 is totally misunderstanding.
And I update this PR after the comment.

knorth55:
Finally, this PR contains several modification.
- change the distance in `auto-dock` function
- do `clear-costmap` in every `go-to-spot` function
- pass `dock` position to improve `auto-dock` success ratio.

the comments posted in this thread are totally misunderstood about this PR, and I changed this PR after the comments.
I have no idea what the policy is, but I fix to work `auto-dock` properly.
It is nonsense to talk about coding policy in this PR.
I'm using this commit in Fetch15 now for go-to-kitchen app.

knorth55:

PR #1172, opened 2019-11-13
[fetcheus] skip undocking when subscribe timeout in go-to-spot

knorth55:
Because of message type difference between melodic and indigo, `go-to-spot` stacks in subscription of `/battery_state`.
I change to skip the subscription in `go-to-spot` when the subscription is timeout.


708yamaguchi:
Thank you.

In order to inform users of the purpose of your code, Please add a comment like
```
;; Because of message type difference between melodic and indigo
```

knorth55:
Thx, I updated to add comment.

knorth55:
@k-okada I want to use this for B3's seminor. Can you review and merge this?

k-okada:
@knorth55 please check https://github.com/jsk-ros-pkg/jsk_roseus/pull/631 whether if we can rewrite this code with `AnyMsg`

PR #1175, opened 2019-11-15
[jsk_fetch_accessories] add fetch_head_box v1

sktometometo:
added STL and STEP design files of fetch_head_box v1


k-okada:
memo : this is just CAD data, not included from fetch urdf model.
c.f.  https://github.com/jsk-ros-pkg/jsk_robot/pull/1186

PR #1143, opened 2019-10-04
[WIP] Merge PR1040 branch to master

knorth55:
move from #1081 
this PR is now split into PRs below;

## Main changes

- [ ] #1342
- [ ] #1417
- [ ] #1491 

## New package dependency
- https://github.com/knorth55/app_manager_utils
- https://github.com/knorth55/coral_usb_ros

## Merged
- `jsk-ros-pkg/jsk_robot`
  - [x] #1121
  - [x] #1136
  - [x] #1141
  - [x] #1142
  - [x] #1144
  - [x] #1150
  - [x] #1178
  - [x] #1180
  - [x] #1182
  - [x] #1183
  - [x] #1184
  - [x] #1191
  - [x] #1192
  - [x] #1198
  - [x] #1223
  - [x] #1225
  - [x] #1227
  - [x] #1233
  - [x] #1234
  - [x] #1241
  - [x] #1247
  - [x] #1248
  - [x] #1255
  - [x] #1256
  - [x] #1258
  - [x] #1281
  - [x] #1282
  - [x] #1284
  - [x] #1286
  - [x] #1300
  - [x] #1308
  - [x] #1310
  - [x] #1314
  - [x] #1340
  - [x] #1341
  - [x] #1347
  - [x] #1359
  - [x] #1360
  - [x] #1361
  - [x] #1381
  - [x] #1392
  - [x] #1393
  - [x] #1401
  - [x] #1418
- `PR2/app_manager`
  - [x] https://github.com/PR2/app_manager/pull/25
  - [x] https://github.com/PR2/app_manager/pull/26
- `tork-a/roswww`
  - [x] https://github.com/tork-a/roswww/pull/51

cc. @YutoUchimi @Affonso-Gui @Kanazawanaoaki

PR #1178, opened 2019-11-18
Update parameters for navigation / safe_teleop

YutoUchimi:
Split from #1143 

Summary of commits by @furushchev 
- Use 2d costmap: VoxelLayer -> ObstacleLayer.
- Fix laser topics used for obstacle observation in move_base.
- Adjust params for move base tilt observation persistence.
- Update safe teleop params.

Summary of commits by @iory 
- Add option to use teb_local_planner.

Summary of commits by @knorth55 
- Use move_base local costmap parameter for safe teleop base.

Summary of commits by @YutoUchimi 
- Use pentagonal footstep polygon.
- Use `omni-corrected` model type and set valid parameters for that type.
- Update amcl parameters for robust localization against wheel slipping.
- Limit `max_vel_y` to avoid running like a crab.
- Decrease `sim_time` and increase `controller_frequency` to avoid frequent navigation chattering on real robot which occurs when forward simulation is not finished in the timeout.
- Increase `min_obstacle_height` so as not to observe some noised points on floor plane from tilt laser.
- Increase padding of tilt_laser_self_filter not to regard PR2's links as obstacle while moving base.
- Explicitly disable recovery_behavior which is hard to notice because `max_planning_retries` is set to -1 in default, which actually disables recovery_behavior.
- Update params for cost calculation used during global planning.
- Do not allow to plan to go through unknown space.
- Allow small translational movement when rotating in place to enable to move in '-x' direction.


YutoUchimi:
I chery-picked b58bd64 by mistake.
It should be merged into #1121 , so I reverted that commit from this PR.

k-okada:
please check on fridge demo with @takayuki5168

takayuki5168:
@k-okada  I checked fridge demo with @HiroIshida 6 times last week. There are some below failures, but anything below has nothing to do with this PR.
- Tilt laser device was not recognized.
    - After rebooting PR2 by circuit breaker, the device was found, but costmap is abnormal. The region in front of PR2 is light blue, which means PR2 could not move at all.
    - After recalibrating tilt laser twice, the costmap becomes normal, and everything worked with demo.
- TF was not published. After rebooting  PR2, everything worked with demo.

PR #1180, opened 2019-11-18
Fix typo in kinect_head_c2.launch for PR2

YutoUchimi:
Split from #1143


k-okada:
is this typo?

knorth55:
yes, it is.
this is working on pr1040.

PR #1183, opened 2019-11-18
Add automated_logger

YutoUchimi:
Split from #1143 

Note: automated_logger.py is added in this PR, but not used.


k-okada:
please use this in demo (sandwitch / elevetors) @knorth55

knorth55:
I have no idea about this PR...
No one knows what automated_logger is and how to use automated_logger.
It may take long time to analyze this code and use this node correctly.

knorth55:
move to 1360

PR #1184, opened 2019-11-18
Fix quater point cloud

Affonso-Gui:
Know that `quater/points` cloud? It happens that it is not actually a quarter, but half.

(since https://github.com/jsk-ros-pkg/jsk_robot/commit/2e9ae29eb62cebe99448333f52a6d4f1bc21d79d)

@knorth55 
@YutoUchimi


knorth55:
Good catch!

knorth55:
I cherry-picked this commit to PR1040 branch #1143

k-okada:
please check with demo (sandwitch / elevetors)

Affonso-Gui:
@k-okada I believe none of them is using quater cloud, but will double check

Affonso-Gui:
@k-okada Neither the refrigerator nor the subway nor any other program in jsk_demos seem to be using the quater points.

Have been already testing on real robot for 1 week without problems. 

Also, testing with the recognition node also show that the quater processing is stable and faster than the half one (as it should be).

k-okada:
PR2 demo is not working on Friday, please check it
--
◉ Kei Okada



2019年11月25日(月) 18:19 Affonso, Guilherme <notifications@github.com>:

Affonso-Gui:
PR2 demo was caused by costmap problems (@takayuki5168) and is unrelated to this. (pr2_move_base has no references to quater points as well)

We tested the demos once again now, and were all moving fine. We can keep up the testing for a couple more weeks if necessary.

knorth55:
I don't know that demo on Friday is failed.
However, this PR is not related to the demo failure, I strongly agree.
No one used quater point cloud so far, and Guiga wanted to use the topic and found this bug.
I think he made a good job to find out this bug.
Now we are using PR1040 with this commit, and Guiga is using this topic in his demo.
So, let's run your demo again and again and prove that this topic is useful @Affonso-Gui

takayuki5168:
@k-okada  I checked fridge demo with @HiroIshida 6 times last week. There are some below failures, but anything below has nothing to do with this PR.
- Tilt laser device was not recognized.
    - After rebooting PR2 by circuit breaker, the device was found, but costmap is abnormal. The cost in front of PR2 is light blue, which means PR2 could not move at all.
    - After recalibrating tilt laser twice, the costmap becomes normal, and everything worked with demo. So we think that the calibration of tilt laser became shifted as time passed by.
- TF was not published. After rebooting  PR2, everything worked with demo.

PR #1188, opened 2019-12-07
Add fetch virtual joint for fetch moveit config

knorth55:
add virtual joint from fetch to map for adding collision objects from map frame


708yamaguchi:
I am sorry, but I cannot understand what you are doing.
Could you please explain the details?

Is robot fetch?
https://github.com/jsk-ros-pkg/jsk_robot/pull/1188/files#diff-acd4fe7d4341f28eb43599e661425bd7R2

knorth55:
Thanks, I fixed the typo: `baxter` -> `fetch`

knorth55:
without this PR, collision object moves when fetch moves.
with this PR, collision objects are associated with map frame, so that objects do not move when fetch moves.

knorth55:
but be careful that all collision objects are associated to map frame, so you cannot add the collisiob object associated to robot base frame as easily as before.
maybe we need to support adding collision object with specific frame in moveit for the future.

PR #1190, opened 2019-12-17
[DO NOT MERGE THIS] Device monitor

takayuki5168:
Recently there sometimes happens that base_laser device cannot be recognized by ROS node.

By conversation with Mr.Uchimi that we should monitor that device, I wrote a program of monitoring devices including base_laser.

I will check this program works well with PR1040 and refine.


takayuki5168:
Advised by Prof.Okada, I will utilize /diagnostics topic or sanity_check to make a system general with several robots.

takayuki5168:
I will utilize TopicPublishedChecker of [sanity_lib.py](https://github.com/jsk-ros-pkg/jsk_common/blob/master/jsk_tools/src/jsk_tools/sanity_lib.py)

takayuki5168:
I will add function of monitorTopicIsPublished in [sanity_lib.py](https://github.com/jsk-ros-pkg/jsk_common/blob/master/jsk_tools/src/jsk_tools/sanity_lib.py).

With checkTopicIsPublished, we can check the topic once, but I want to monitor (check always) the topic, so I think we need monitorTopicIsPublished.

I will implement that we can set callback function in monitorTopicIsPublished which is called when the topic is death. I want to hear your opinion before implementation. @k-okada

takayuki5168:
I don't do maintenance on this PR anymore, so I close.
If this PR should still be opened, please make another PR.

PR #1191, opened 2019-12-20
Add service interface to look-forward to start/stop from another process

YutoUchimi:
I added  a service interface of start/stop to look-forward-in-nav.l in order to move head from another process without conflicting.


k-okada:
Nice , how about  showing current status (`*look-enable*`). For example https://github.com/jsk-ros-pkg/jsk_robot/blob/f6e9e1055a80679cf49a52c97a1988500ab78bee/jsk_pr2_robot/jsk_pr2_startup/jsk_pr2_move_base/look-forward-in-nav.l#L44

YutoUchimi:
That's good idea.
I added that info.

PR #1192, opened 2019-12-20
Enable recovery_behavior

YutoUchimi:
Change parameters to enable recovery_behavior.
This can be useful when robot is stuck by obstacles.

Still testing with real robot.


k-okada:
Why you disabled at https://github.com/jsk-ros-pkg/jsk_robot/commit/60f8a5a8bbec218f9545b85f03d5ce508becf81a#diff-b7c5d42e85d8c0ca16a95d0eafab7b60 and re-enabled?

YutoUchimi:
- Why I disabled?
According to http://wiki.ros.org/move_base#Parameters , the default value of `~max_planning_retries` is `-1`, which means recovery_behavior is disabled by default.
To notice this fact easily, I added a comment in b51f2c6e0fd4b47bad8e037d0121d3eb2dfba914 .

- Why I (re-)enabled?
The problem is how to deal with moving obstacles such as human.
Assume that human is standing 2 meters in front of the robot, and robot rotates 180 degrees in place.
Then a user commands the robot to go to the place the human was standing.
In this case, even if the human went away, the robot fails to plan to reach the goal, since the robot cannot observe that the human has already gone.
The robot has to rotate in place to update costmap, so I enabled recovery_behavior.

With this PR, move_base itself will give up on going to the target spot in some case.
If you want to retry forever, you must control from upper level such as euslisp `(send *ri* :move-to coords :retry *inf*)`

PR #1193, opened 2020-01-07
[jsk_fetch_startup] use plugin for applist load in fetch

knorth55:
use plugin for applist load in fetch as PR2


knorth55:
now i'm testing on Fetch15

knorth55:
yes you are right. we do not need `applist` arg if `use_applist` is `false`.

k-okada:
If I understand correctly, this change load list of apps from `jsk_fetch_startapp/apps`  to `export` tag. Thus increase the number of apps listed on the main menu. If this is the case.
Please add information on the list of apps before/after this change

knorth55:
So far, all fetch app is located in `jsk_fetch_startup/apps`, so apps in web menu does not increase.
however, if we add new fetch apps in some repo (like 2019 semi demo in jsk_demos), the new app will appear in web menu.
OK, so I will make 2019 semi program into app.

k-okada:
I see , can we write something like `platform: *` ? if so, can we fix
hello_world.apps ?
when ever you change the code, you'd better to show something that can not
achieve without this change.
--
◉ Kei Okada



2020年1月9日(木) 12:40 Shingo Kitagawa <notifications@github.com>:

knorth55:
https://github.com/PR2/app_manager/blob/kinetic-devel/src/app_manager/app_list.py#L131-L136
according to this line, even if we set `platform: *`, app will be filtered in these lines.
but it is good that we can use `platform: *` or `platform: all`.

knorth55:
app like `time_signal` in #1194 , we can run this app in all robot.

knorth55:
before this PR, all app should be located in `jsk_fetch_startup/apps`.
For example,  `welcome_to_jsk` app in https://github.com/jsk-ros-pkg/jsk_robot/blob/master/jsk_fetch_robot/jsk_fetch_startup/apps/welcome_to_jsk/welcome_to_jsk.xml is located in `jsk_fetch_startup/apps`, but all the source is located in `jsk_demos`.
with this PR, we can avoid this by adding fetch apps in `jsk_demos`.

k-okada:
it is always better to start with something already exists, before
writing new code. whey not change
https://github.com/jsk-ros-pkg/jsk_robot/tree/master/jsk_pr2_robot/jsk_pr2_startup/apps/hello_world
and others to platform: all?

--
◉ Kei Okada

2020年1月9日(木) 18:19 Shingo Kitagawa <notifications@github.com>:

knorth55:
this PR is not related to `platform: all` problem.
this PR is just loading apps from all repo not only from `jsk_fetch_startup`.

knorth55:
I made PR to add `all` platform https://github.com/PR2/app_manager/pull/17

PR #1200, opened 2020-02-17
[fetch] make jsk_fetch_accessories to ros package

sktometometo:
- changed the jsk_fetch_accessories directory to a ros package so that mesh files in the dir can be loaded from URDF file like https://github.com/sktometometo/fetch_ros/blob/add-fetch-base-camera-mount-mk1/fetch_description/robots/fetch.urdf#L41


jsk_fetch_robot/jsk_fetch_accessories/package.xml
```diff
@@ -0,0 +1,13 @@
+<package>
+  <name>jsk_fetch_accessories</name>
+  <version>1.0.0</version>
```
pazeshun: Please set the same version as other packages (1.1.0?)

pazeshun:
Should be tested well

PR #1201, opened 2020-02-18
[fetch] add monitor fixture to guard side of monitor

708yamaguchi:
I add monitor fixture to guard side of the monitor.

fixture

fixture_side


knorth55:
Good.
For further improvement, it is better to think about ordering this part to some company, not making by 3D printer.

708yamaguchi:
Thank you.

I will watch Fetch with these new parts for a few days.
If there is no problem, I want to outsource these parts after adding some improvements.

708yamaguchi:
I added the imprint "fetch" to the left side of the part and spray-painted the fetch letters blue.

monitor_fixture_cad
monitor_fixture1
monitor_fixture2

CC @mmwise

708yamaguchi:
I think this is OK to be merged.

PR #1207, opened 2020-02-27
[WIP] Add etc files

sktometometo:
- add upstart scripts and launch files which are usually in /etc/init/ and /etc/ros/indigo.
- add a script to deploy the upstart scripts
- modify the upstart scripts so that they will use jsk_fetch_startup version of launch files instead of robot.launch in /etc/ros/indigo


sktometometo:
wrongly create a PR.

PR #1215, opened 2020-04-13
[wip] add dockerfile for docker build

knorth55:
with this dockerfile and hook, we can set autonomous docker build on dockerhub.


knorth55:
I want this dockerfiles to be merged and set build trigger in dockerhub for each commit.
This build trigger may be useful for debugging the repository.

Issue #1216, opened 2020-04-18
run fetch-interface.l without roscore


takayuki5168:
There is an initialization of a ros node in [`fetch-interface.l`](https://github.com/jsk-ros-pkg/jsk_robot/blob/95959dc105b2d2c90d1d5d1bad49cd2a6560aa21/jsk_fetch_robot/fetcheus/fetch-interface.l#L1), so we cannot run `fetch-interface.l` without `roscore`.
Are there any reasons to initialize the ros node?

In `pr2-interface.l`, there is no initialization of ros node, which I think much more convenient.

knorth55:
maybe not.
can you run fetch in kinematics simulator without ros?
if so, it is better to remove the initialization.

takayuki5168:
@knorth55 
Kinematics simulator means using `robot-interface` right?
We cannot generate an instance of `robot-interface` without ros because it uses ros publisher or subscriber. So when we use `robot-interface`, we have to run roscore.

knorth55:
so, why do you want to remove the line?

takayuki5168:
It is convenient when we only generate motion in euslisp, which is independent of ros,..., but I found that when we want to do that, we could realize by loading not "pr2-interface.l" or "fetch-interface.l" but "pr2.l" or "fetch.l".

However, we should unify whether we initialize the ros node or not in "pr2-interface.l" and "fetch-interface.l". I think both is fine.

Affonso-Gui:
I do agree on moving the node initialization to `fetch-init` in resemble to the pr2-interface

Not very relevant, but this can cause misleading behavior in cases like the below example, in which the node would be initialized as "fetch" rather than as "my_node"
```
(load "package://fetcheus/fetch-robot.l")
(ros::roseus "my_node")
(fetch-init)
```

Also, pr2 uses "pr2_eus_interface" and fetch uses "fetch". We could unify those as well.

takayuki5168:
@Affonso-Gui 
Thanks for your favorable comment.

In `fetch-interface.l` the ros node is about to be initialized again as `fetch_eus_interface`, but ignored because the ros node has been already initialized as `fetch` in the first line.

I made a simple PR for this. https://github.com/jsk-ros-pkg/jsk_robot/pull/1218

knorth55:
IMO, if it causes some error, we need to remove it, but it not, it is personal preference.
I prefer removing the line, but I don't strongly recommend it.

[...]

PR #1218, opened 2020-04-21
[fetcheus] remove unnecessary ros node initialization in the first line

takayuki5168:
For https://github.com/jsk-ros-pkg/jsk_robot/issues/1216


Affonso-Gui:
The original `(ros::roseus "fetch")` initialization seems to be present since the first commit at https://github.com/jsk-ros-pkg/jsk_robot/commit/d77ffcb5937754345deb99bddc876e280f738b44#diff-d7eb7619176f5641c713e05e846088d4

@k-okada Any particular reason you can remember of for initializing the node when loading instead of from within `fetch-init` ? If not I think this PR is valid and might be a nice idea to integrate it before we have several fetch demos with the both machines that can possibly depend on the node name

@708yamaguchi @knorth55

knorth55:
how about just removing this, but change like `pr2-interface.l`?
https://github.com/jsk-ros-pkg/jsk_pr2eus/blob/master/pr2eus/pr2-interface.l#L644

Affonso-Gui:
@knorth55 

What do you mean? `fetch-init` is already like the `pr2-interface.l` one:
https://github.com/takayuki5168/jsk_robot/blob/remove-first-ros-node-initialization/jsk_fetch_robot/fetcheus/fetch-interface.l#L334

Or are you saying that we should change the `pr2-init` syntax to something else?

knorth55:
@Affonso-Gui Oh, I didn't know that.

knorth55 (approved):

PR #1219, opened 2020-05-19
[jsk_fetch_startup] add rgb compressed throttle in fetch

knorth55:
i add `/head_camera/rgb/throttled/image_rect_color/compressed`, which does not exist now.


708yamaguchi:
It seems good to me.

There exists `jsk_topic_tools/LightweightThrottle` node for compressed depth image, but not for rgb image.
https://github.com/jsk-ros-pkg/jsk_robot/blob/fcc3d3909a566f60249db1e3910147ec6a0d1a14/jsk_fetch_robot/jsk_fetch_startup/launch/fetch_sensors.xml#L63-L71

PR #1222, opened 2020-06-15
[jsk_pr2_accessories] Add fingertips used for @AnneKoepken's demos

pazeshun:
cc @Kanazawanaoaki 

This PR includes STL files of the fingertips used for @AnneKoepken's demos.
Unfortunately, she used Shapr3D as 3D CAD (maybe no one in JSK uses this) and she lost the original data (.shapr) of the left hand fingertip.


pazeshun:
Thank you!

PR #1223, opened 2020-06-19
[jsk_pr2_startup] add throttle for tilt_scan and compressed rgb/depth images

knorth55:
i update pr2 launch to get throttle topics below
this is for rosbag record.

- `/tilt_scan_throttled`
  - throttle for `/tilt_scan`
- `/kinect_head/rgb/throttled/image_rect_color/compressed`
  - throttle for `/kinect_head/rgb/image_rect_color/compressed`
- `/kinect_head/depth_registered/throttled/image_rect/compressedDepth`
  - throttle for `/kinect_head/depth_registered/image_rect/compressedDepth`


knorth55:
I'm testing this on PR1040.

jsk_pr2_robot/jsk_pr2_startup/jsk_pr2_sensors/kinect_head.launch
```diff
@@ -85,6 +85,15 @@
         <remap from="~output" to="throttled/image_rect_color" />
         <param name="update_rate" value="$(arg throttle_rate)" />
       </node>
+      <node name="throttle_rgb_compressed"
+            pkg="nodelet" type="nodelet"
+            args="load jsk_topic_tools/LightweightThrottle /$(arg camera)/$(arg manager)"
```
iory, line 90: I suppose using `jsk_topic_tools/SynchronizedThrottle` to publish synchronized RGBG images.

iory: also camerainfos

knorth55: I don't know which is better.
Fetch and PR2 uses `LightweightThrottle` now, so we should try `LightweightThrottle` first,
and we can switch to `SynchronizedThrottle` when it is needed.

iory: I think that these patches use-cases are as follows.
- Visualize RGB-D images at remote PC
- Visualize PointCloud at remote PC. (Should be (a)synchronized)
- Run recognition program with the throttled image to reduce the resource of PC. (Should be (a)synchronized)

Considering the above, I think `SynchronizedThrottle` is better.
https://github.com/jsk-ros-pkg/jsk_hsr/blob/master/jsk_hsr_startup/launch/hsr_bringup.launch#L458-L472

> we can switch to SynchronizedThrottle when it is needed.

However, if you don't think you should do this right away, this change may be a good start.

knorth55: if it is so, let me open another PR to use `SychronizedThrottle`.
this PR purpose is to add missing throttle, not improving others.

iory: What is this PR's purpose? Please add the comment.

iory: Purpose of throttled image.

iory: for beginners

knorth55: throttle image is for rosbag.

iory: Isn't it better to write down the reason why you should use throttle images at the top comment of this PR?

knorth55: I added.

iory:
@k-okada 
This patch does not break backward compatibility, just add the throttle node.
It's easy to merge.

PR #1220, opened 2020-05-21
[jsk_fetch_startup] add rosbag_play.launch and rosbag_record.launch

knorth55:
originally from @sktometometo 
modified by @knorth55 

I add `rosbag_play.launch` and `rosbag_record.launch` for fetch.
in that point, we add `latched_topic_publisher.py` node for `/tf_static` and `/map` topics.

PR #1228, opened 2020-07-01
[jsk_pr2_startup] add rosbag_play.launch and rosbag_record.launch for pr2

knorth55:
add rosbag_play.launch and rosbag_record.launch as fetch #1220 
you can replay rosbag as fetch with this PR.

pr2_pick_rviz


knorth55:
cc. @Kanazawanaoaki

knorth55:
@iory 
I think it is a bit difficult to find this launch file in `jsk_data`.
I think `jsk_robot` is the right place to have this launch.

iory:
> I think it is a bit difficult to find this launch file in jsk_data.

If so, we need to write `README` that contains how to `record` and `play` rosbag to find those launch files.

PR #1226, opened 2020-06-25
[jsk_fetch_startup] switch to use jsk_maps multi_map_server

knorth55:
depends on https://github.com/fetchrobotics/fetch_ros/pull/144 and https://github.com/fetchrobotics/fetch_ros/pull/146

this PR switch to use `jsk_maps` `multi_map_server`.
this is useful for move_base with elevators.

we are testing this PR on Fetch15.
Fetch15 could go to 3rd floor with joy teleop.

cc. @708yamaguchi

PR #1231, opened 2020-07-07
 [jsk_fetch_startup] use keepout maps 

knorth55:
~merge after #1226~
~depends on https://github.com/jsk-ros-pkg/jsk_demos/pull/1300~

with this PR, fetch uses keepout for entire eng2 building.

cc. @708yamaguchi


708yamaguchi:
Today, we have realized that it is important to publish `/map` (for amcl) and `/map_keepout` (for move_base) separately.
Therefore, we need to launch `start_map_$(arg map_frame).launch` twice, for `/map` and `/map_keepout`.

However, in that case, the map topic which is now published from `fetch_nav.launch` must not be published. 
We may need to change `fetch_nav.launch` so that we can select whether to publish map topic.

knorth55:
@708yamaguchi 
> Today, we have realized that it is important to publish /map (for amcl) and /map_keepout (for move_base) separately.
Therefore, we need to launch start_map_$(arg map_frame).launch twice, for /map and /map_keepout.

You are right. We need to figure it out.

>However, in that case, the map topic which is now published from fetch_nav.launch must not be published.
We may need to change fetch_nav.launch so that we can select whether to publish map topic.

Is this what you mean? https://github.com/fetchrobotics/fetch_ros/pull/146

708yamaguchi:
> Is this what you mean? fetchrobotics/fetch_ros#146

Sorry. I looked at https://github.com/fetchrobotics/fetch_ros/blob/melodic-devel/fetch_navigation/launch/fetch_nav.launch

I mean
- if `use_keepout` is true, `/map_keepout` is published from both map_server and multi_map_server.
- if `use_keepout` is false, fetch's `move_base` subscribes `/map`, not `/map_keepout`

I think both cases have problem.

knorth55:
IMO,
so, first, we need to modify this PR to publish both `/map` and `/map_keepout`.
then set `use_keepout:=true` and `launch_map_server:=false` in `fetch_nav.launch`.

708yamaguchi:
Yes, that's right. I'm sorry for confusion.

knorth55:
@708yamaguchi thanks, I update this PR 6d604efbaf30d7b24b3fadd6238bc1ff452176a4 and jsk_maps [`9c7c5d0` (#1300)](https://github.com/jsk-ros-pkg/jsk_demos/pull/1300/commits/9c7c5d0eed11f5f24c97e9f925e4761b0339ce3f)
[`a7e0cc7` (#1300)](https://github.com/jsk-ros-pkg/jsk_demos/pull/1300/commits/a7e0cc7b6afb5cd8f396b5a841469e2a7ee6bcd8)

I checked with Fetch15 and he can go from 73B2 to the front of elevator at 7F.

knorth55:
@708yamaguchi 

>One thing is that we need to switch both /map and /map_keepout when we run (change-floor) in euslisp.

Actually, we don't need to consider about this.
Here is the reason.
In order to switch map, `initialpose3d` uses `/map_reload` topic
https://github.com/jsk-ros-pkg/jsk_demos/blob/master/jsk_maps/tools/initialpose3d.py#L55

We add `map_server` and `map_keepout_server` (`multi_map_server`) for `/map` and `/map_keepout`,
and both nodes subscribe `/map_reload`.
So, we can switch both map simultaneously by publishing a topic to `/map_reload`.
https://github.com/jsk-ros-pkg/jsk_common/blob/master/jsk_ros_patch/multi_map_server/src/main.cpp#L244

```
fetch@fetch15:~/ros/indigo/src/jsk-ros-pkg/jsk_demos/jsk_maps/src$ rostopic info /map_reload
Type: std_msgs/String

Publishers:
 * /initialpose3d (http://fetch15:33113/)

Subscribers:
 * /map_keepout_server (http://fetch15:52138/)
 * /map_server (http://fetch15:53926/)
```

Therefore, now Fetch15 uses `/map` for `amcl` and `/map_keepout` for `move_base`, and he can switch both map by `(change-floor "/eng2/3f :base-frame-name "base_link")`.

fetch_change_floor_keepout

708yamaguchi:
Thank you for your explanation. I understand the role of the two `multi_map_server` nodes.
Great work for this multi-floor and keepout map project! :100:

knorth55:

PR #1232, opened 2020-07-14
[wait for upstream] [jsk_fetch_startup] add basic authentication in roswww on fetch

knorth55:
wait for upstream
depend on https://github.com/tork-a/roswww/pull/51

add basic authentication for all roswww pages

roswww_basic_auth

cc. @708yamaguchi


708yamaguchi:
I accessed this page with chrome, but at first topic list was not displayed.
It behaves correctly after two reloads. (I don't know the reason, but 2 reload seems important)

rwt_error

I use chrome browser.
When I use normal window, this only happens once.
When I create new incognito window, this happens repeatedly.

knorth55:
@708yamaguchi I confirmed the issue, but I can get proper page at once when I connect to pr1040.
I checked with web developer and some css are failed to download.
maybe something got timeout.

708yamaguchi:
Thank you for thoughts. I agree with you.
It seems that not only css files but also ROSLIB or jQuery are sometimes not loaded.

The error says,
```
 Failed to load resource: net::ERR_CONTENT_LENGTH_MISMATCH roslib.js:1
```
This may be cache problem (?)
https://cpoint-lab.co.jp/article/201809/5517/

k-okada (commented):
do we need to add `tork-a/roswww`
https://github.com/jsk-ros-pkg/jsk_robot/blob/master/jsk_fetch_robot/jsk_fetch.rosinstall ?

knorth55:
yes we need https://github.com/tork-a/roswww/pull/51.
@k-okada do you have a plan to merge https://github.com/tork-a/roswww/pull/51 ?

PR #1195, opened 2020-01-25
[jsk_robot_startup/util/mux_selector] Add wait option for topic

iory:
I added a wait option to wait for the topic of other nodes that are up at the same time.
- [x] Add `wait` option
- [x] Add docs

PR #1234, opened 2020-07-15
[merge after #1195, #1237 and #1260] Common safe teleop for JSK mobile robots

708yamaguchi:
This PR depends on #1195 and #1237 

In this pull request, I did
- create common safe teleop system for JSK mobile robots
  - This system allows teleop inputs other than joystick. So, for example, fetch will be able to move with keyboard input.
- apply the system to fetch and PR2
  - This system currently runs on fetch and PR2.
  - For the time being, I'll keep an eye on the behavior.
- update README
  - I update documentation
  - I add system diagram

System diagram (UPDATED in 2020/10/07):
jsk_safe_teleop_system

cc @knorth55


knorth55:
We are testing this PR on Fetch15 and PR1040.

jsk_robot_common/jsk_robot_startup/util/mux_selector.py
```diff
@@ -51,12 +51,11 @@ def add_trigger(topic, expr, select, index):
...
     if(topic_type is None):
-        rospy.loginfo('%s is not published yet', topic)
-        return None
+        rospy.logwarn('%s is not published yet', topic)
 
     if(topic_class is None):
-        rospy.loginfo('%s is not builded yet', topic_type)
-        return None
+        rospy.logwarn('%s is not builded yet', topic_type)
+        topic_class = rospy.AnyMsg
```
iory, line 58: Please explain the purpose of this line.

708yamaguchi: Thank you for your review. 

Without this pull request,
1. I want to use `topic_tools/relay` to use the same topic as the system diagram
2. The type of `topic_tools/relay` output topic is not defined when the teleop system starts.
3. `mux_selector.py` cannot subscribe the output topic
https://github.com/jsk-ros-pkg/jsk_robot/blob/b7163007df321d3e9349cbfe1bb471bb163056fd/jsk_robot_common/jsk_robot_startup/util/mux_selector.py#L64

Therefore, I use `rospy.AnyMsg` if the topic type is not defined.

iory: I see.
Did you see the https://github.com/jsk-ros-pkg/jsk_robot/pull/1195 ?
You can cope with the problem with its `wait` option.

708yamaguchi: Thank you for your advice.
I did not know https://github.com/jsk-ros-pkg/jsk_robot/pull/1195, but now I think https://github.com/jsk-ros-pkg/jsk_robot/pull/1195 is useful for this case.

I pushed additional commits.

708yamaguchi:
According to @knorth55 's advice, I update the system.
In the new system,
- I do not use several `safe_teleop_base` node to reduce CPU load.
- I use different vel_type_selector for fetch and PR2, because fetch does not use `/base_controller/cmd_vel` topic.
(If `mux_selector.py` is used with unknown type topics and `wait=True`, `mux_selector.py` does not work. So we need to use different vel_type_selector node)

JSK teleop_base system

knorth55:
This PR seems good, and we are testing this on PR1040 and Fetch15.
With this PR, all navigation system is integrated in PR2 and Fetch, thus we can easily move both robot with same topic name space.

knorth55:
We found that vel_type_selector select default topic name often, so we change default topic to previously selected topic.

jsk_safe_teleop_system

708yamaguchi:
Thank you very much for your help!

@mqcmd196

knorth55 (approved):

708yamaguchi:
I add arg `use_safe_teleop`.
I need this arg when I want to use mux system in this launch but don't want to use safe teleop node because safe teleop node increases CPU load.

CC @mqcmd196

PR #1235, opened 2020-07-24
[mux_selector.py] Add rosparam to set type of rostopic

708yamaguchi:
In this pull request, I add rosparam (`~topics`) to set type of rostopic.
With this feature, we can use unknown type topic, such as output topic of `topic_tools/relay`.

@knorth55


708yamaguchi:
Thank you for your advice, but I create wrong branch `jsk-ros-pkg/add-type` so I close this PR and reopen.

PR #1237, opened 2020-07-24
[jsk_robot_startup/mux_selector.py] Add rosparam to set type of rostopic

708yamaguchi:
Rewrite version of #1235 . Thank you  for your review @knorth55.

In this pull request, I add rosparam (~topics) to set type of rostopic.
With this feature, we can use unknown type topic, such as output topic of topic_tools/relay.


knorth55:
I close this PR because #1234 is merged.

PR #1239, opened 2020-08-05
[jsk_fetch_startup] add use_build_map in fetch_bringup.launch

knorth55:
add `use_build_map` argument for building map
this change is originally made by @sktometometo 

cc. @sktometometo @708yamaguchi


k-okada (changes requested):
@sktometometo need to resolve conflicts
Are we using this ? @708yamaguchi

sktometometo:
@k-okada We still need this if we build a map.

PR #1227, opened 2020-06-26
[merge after #1225] add rwt_image_view for pr2

knorth55:
depends on #1225 

add `rwt_image_view` for pr2.
you can see image topic from web browsers.

PR #1247, opened 2020-09-17
[merge after #1227] [jsk_pr2_startup] launch roswww, rosbridge, web_video_server on c2

knorth55:
merge after #1227 

now PR1040 launches roswww, rosbridge, web_video_server on c2 to reduce load on c1.

cc. @ishiguroJSK


k-okada (commented):
@knorth55 I have resolved the conflict, please check if it is correct.

knorth55:
yes, I think it's ok.

PR #1251, opened 2020-09-29
[jsk_pr2_startup] add virtual force launch

Kanazawanaoaki:
I tried to publish virtual force of both hands with `jsk_pr2_startup/pr2_bringup.launch`.  
It publish `/left_endeffector/wrench` and `/right_endeffector/wrench`.

I didn't know what to do,but this is calling the [virtual_force_publisher.cpp](https://github.com/jsk-ros-pkg/jsk_common/blob/master/virtual_force_publisher/src/virtual_force_publisher.cpp ) in `jsk_common`.


knorth55:
Good, but can we include the files in https://github.com/jsk-ros-pkg/jsk_common/pull/1656 ?
it would be smarter, I think.

Kanazawanaoaki:
I see. I made it include the launch file.  https://github.com/jsk-ros-pkg/jsk_robot/pull/1251/commits/f4962d29260059a6d5905da99970775b46301efb

PR #1252, opened 2020-09-30
[jsk_fetch_startup] Make the Rviz navigation path thicker

708yamaguchi:
I make the Rviz navigation path thicker.

This makes it easier for us to see the current navigation path (the pink lines in the figures below) from the monitor on the back of fetch.

|Before|After|
|--|--|
|||


knorth55:
it is already merged

PR #1253, opened 2020-10-03
[jsk_fetch_startup] Add network detector to switch services depending on the network

708yamaguchi:
I added the node to detect whether fetch is connecting to the network.
We can check the network condition by `rostopic echo /network/connected`.

As an example of use, I'm going to use google speach_to_text when the network is on and use julius when it's off.

@mqcmd196


knorth55 (approved):

k-okada:
So can we use https://github.com/jsk-ros-pkg/jsk_robot/pull/1285 to detect network status ?

knorth55:
yes, we can us #1285, I think.
if we got `0` or no message from `network_status.py`, we can determine that the network is down.

PR #1260, opened 2020-10-16
[jsk_robot_startup] fix condition in mux_selector.py

knorth55:
fix condition to suppress rosservice call error with `next_topic` `None`.


knorth55:
I close this PR because #1234 is merged

PR #1268, opened 2020-11-06
[fetcheus] make keepout smaller

knorth55:
make keepout smaller
now keepout is a bit large that arm often collides with the virtual keepout object.

fetch_smaller_keepout


708yamaguchi:
I updated the params at https://github.com/knorth55/jsk_robot/pull/42

PR #1271, opened 2020-11-10
[fetcheus] add provide in fetch-interface

knorth55:
add missing `(provide :fetch-interface)` to prevent multiple loading


Affonso-Gui:
I think I already mentioned this in some other issue and I am not saying that it is bad to add this (I support this PR), but just to be very clear `provide` does actually very little to help files from being loaded multiple times in euslisp.

This is because
1. provide is already called from within require
1. you need to make sure that the provide and require share the same module-name

So by adding this provide line in fetch-interface.l the only cases in which you avoid reloading is on:
```lisp
(load "fetch-interface.l")
(require :fetch-interface "fetch-interface.l")  ;; not loaded

(require "fetch-interface.l")
(require :fetch-interface "fetch-interface.l")  ;; not loaded
```

In the following cases it gets reloaded
```lisp
(load "fetch-interface.l")
(load "fetch-interface.l")  ;; loaded

(load "fetch-interface.l")
(require "fetch-interface.l")  ;; loaded

(require :fetch-interface "fetch-interface.l")
(require "fetch-interface.l")  ;; loaded
```

And in the following it wouldn't be reloaded anyways (even without the provide statement)
```lisp
(require :fetch-interface "fetch-interface.l")
(require :fetch-interface "fetch-interface.l")  ;; not loaded
```

Just saying that adding the provide won't do any magic to stop the reloading, and for now the only way to do so is to be careful when loading.
By the way it would be nice to implement a more efficient provide/require interface on euslisp capable of self administrating the module-name tags (no need for the user to control that).

knorth55:
> (load "fetch-interface.l")
(require :fetch-interface "fetch-interface.l")  ;; not loaded

this is also what i wanted to do.

PR #1277, opened 2020-11-18
[jsk_fetch_startup] Add RViz panel to publish /move_base/cancel

708yamaguchi:
I add RViz panel to publish `/move_base/cancel` (Lower left of the figure) and use the same rviz config as knorth55/fetch15.

cancel_panel_

@mqcmd196


mqcmd196:
Thanks! This PR would make our experiments much safer than now.

PR #1287, opened 2020-11-26
[jsk_robot_startup] fix tweet_client_uptime

knorth55:
this PR is splitted from \#1181.
this PR change to use `while` instead of `do-until-key` in `tweet_client_uptime.l`


k-okada (commented):
please describe whey you need to use `while` instead of `do-until-key` and put `ros::rate` within the loop

knorth55:
I don't know why, but I picked up this commit from @YutoUchimi 's PR #1181 .
This change is originally from @furushchev .

k-okada:
is this nesessary?

PR #1292, opened 2020-12-01
[jsk_fetch_startup] add user_speech_notifier in tweet, time_signal, speak_battery apps

knorth55:
speak username in `/app_manager/running_user_name` when the robot starts the app.

related:
https://github.com/tork-a/visualization_rwt/pull/95
https://github.com/knorth55/app_manager_utils/commit/119cb6014c8d1ebce337529f5efe4fcdbe8b7dce


knorth55:
@k-okada with this PR, Fetch speaks who is running the app in `time_signal`, `tweet` and `battery_speak` apps.

708yamaguchi:
Is this program running on Fetch15 or Fetch1075?
I cannot find `/app_manager/running_user_name` by rostopic list and by rosparam list...

knorth55:
it is on fetch15.
can you run `tweet` app from `rwt_app_chooser`?
in that time, you are asked to fill your user name in `rwt_app_chooser`, and fetch15 will say something before starting `tweet` app

708yamaguchi:
Thank you! I understand.
I think this is very nice feature.

Could you add usage or documentation to README.md?

knorth55:
what kind of readme do you want?
readme of `rwt_app_chooser`?
but if user uses `rwt_app_chooser` normally, they will be asked to fill user name.

708yamaguchi:
I think it is better to write below contents to jsk_fetch_startup/README.md
- First, we can register user name for app running.
- We can start apps with empty string, but /app_manager/running_user_name param will not be set.
- Then, we can run an app.
- By username, if one person is using fetch, others cannot use fetch.

knorth55:
@708yamaguchi I have a plan to add `user_warning` node, which the robot alert someone is moving the robot with no `running_user_name` next.
In that time, I will write down the documentation for JSK robot Personal Use app for PR2 and Fetch.

708yamaguchi (approved):
Thank you for your explanation.
I understand.
Sorry for late reply.

PR #1296, opened 2020-12-02
Change tweet app icon

708yamaguchi:
Before
tweet

After
Twitter-Icon


708yamaguchi:
According to @k-okada 's advice, I updated the tweet icon.

tweet

knorth55 (approved):

PR #1318, opened 2021-01-28
Add plug protector stl model for fetch's base station

HiroIshida:
I added stl model and solidworks model of the plug protector for fetch's base station.
 
https://github.com/HiroIshida/jsk_robot/blob/fetch_plug_protector/jsk_fetch_robot/jsk_fetch_accessories/plug_protector/fetch_plug_protector.STL

One must print 2 copies of this part first, and then mount them as in the following pictures. In the latter procedure, one must open the base station. 

plug_support1
plug_support2


708yamaguchi:
Thank you!
Cool!

k-okada (approved):
LGTM

PR #1332, opened 2021-04-21
Add cron install scripts

708yamaguchi:
I add cron scripts for fetch.


708yamaguchi:
Do `catkin clean` after `wstool update`
https://github.com/knorth55/jsk_robot/pull/92

PR #1341, opened 2021-05-31
[jsk_pr2_robot] enabled imu for localization

knorth55:
enable imu for localization.
we checked it does not drift anymore.


k-okada (commented):
@HiroIshida

PR #1347, opened 2021-06-16
[jsk_pr2_lifelog] do not store depth and rgb in mongodb

knorth55:
this PR disable storing rgb and depth image in mongodb.
this is because depth and rgb requires huge storage for mongodb...


k-okada (commented):
@knorth55 are you still sure about this?
if we do no store depth/rgb in mongodb, were do we store image logs.

knorth55:
@k-okada 

> @knorth55 are you still sure about this?
if we do no store depth/rgb in mongodb, were do we store image logs.

it is quite difficult question.
My conclusion is that we should not store rgb or depth images in MongoDB or InfluxDB.
This is because 1) the storage got huge and 2) we cannot easily access the image.
For RGB image, I'm recording into video and audio into `AVI` file with compression, and uploading to google drive.
But for Depth image, I don't have good idea to store it.

PR #1354, opened 2021-07-15
add github action

k-okada:
- add .travis.rosinstall.noetic
- mv jsk_fetch.rosinstall.kinetic -> /jsk_fetch_user.rosinstall.kinetic, see  https://github.com/jsk-ros-pkg/jsk_robot/pull/1008
- add .github/workflows/
- add CHECK_PYTHON3_COMPILE


tongtybj:
May I ask what is the advantage to use github action instead of travis / jenkins?

k-okada:
Since Travis integrate .org (for opensoruce) and .com (for closed source) system as they purchased by  Idera and changed it billing plans, which results in slow build paces at this moment.

To seed up review process, we use Github Action as backup of travis, Github Actions has better response, but (maybe) slow CPU power compared to Travis.

We have not decided whether we should moved to Github Actions, I will Travis back to better OSS support near future.

PR #1206, opened 2020-02-23
[jsk_fetch_startup] fix odom topic of move_base and teleop

knorth55:
this PR depends on https://github.com/fetchrobotics/fetch_ros/pull/137
update odom topic for move_base and teleop to use `odom_combined`.

PR #1362, opened 2021-07-17
Revert "[jsk_fetch_startup] fix odom topic of move_base and teleop"

knorth55:
Reverts jsk-ros-pkg/jsk_robot#1206


knorth55:
we should first revert #1206 because of conflict resolving problems.

knorth55:
move to #1363

PR #1366, opened 2021-07-19
[jsk_fetch_startup] Do not replace sound when other node is speaking

knorth55:
this PR enables fetch to wait speaking until other node is speaking.

cc. @708yamaguchi 

related: https://github.com/ros-drivers/audio_common/pull/173


708yamaguchi (commented):
`jsk_fetch.rosinstall.melodic` and `battery_warning.py` seems good.
But `boot_sound.py` does not reflect https://github.com/jsk-ros-pkg/jsk_robot/pull/1262

I create PR to reflect #1262, and I check this PR today.
https://github.com/knorth55/jsk_robot/pull/130

PR #1370, opened 2021-07-20
[merge after #1208] Add upstart scripts

708yamaguchi:
merge after #1208

I add upstart scripts for fetch.

I cherry-pick the following commits
- 3bff22c ~ dcc0ce2 (for fetch.launch) in https://github.com/knorth55/jsk_robot/commits/fetch15/jsk_fetch_robot/jsk_fetch_startup/launch/fetch.launch
- b961a7a ~ 94b4874 (for upstart_scripts) in https://github.com/knorth55/jsk_robot/commits/fetch15/jsk_fetch_robot/jsk_fetch_startup/upstart_scripts
- 37d27ea ~ b1b2388 (for install_upstart.sh) in https://github.com/knorth55/jsk_robot/commits/fetch15/jsk_fetch_robot/jsk_fetch_startup/config/install_upstart.sh

PR #1328, opened 2021-04-05
[merge after #1208 and #1371] Fetch melodic and new hardware support

knorth55:
This PR supports Fetch-JSK support (Melodic + JSK hardware)

this PR depends on #1208 and #1371, please merge after #1208 and #1371

Thank you @sktometometo @708yamaguchi

PR #1384, opened 2021-09-19
[jsk_fetch_startup] disable base_camera_mount in upstart_scripts/robot.conf

knorth55:
move from #1370 

set `base_camera_mount` false in robot.conf

cc. @sktometometo @708yamaguchi


knorth55:
this is moved to #1328

PR #1374, opened 2021-08-01
Send email if battery is low

708yamaguchi:
Cherry-pick https://github.com/knorth55/jsk_robot/pull/135

fetch sends email if battery is low.

address yaml file is like below:
```bash
$ cat /var/lib/robot/battery_warning_address.yaml
sender_address: sender@hoge.com
receiver_address: receiver@hoge.com
```

The output email is like below:
```
Fetch is low battery
===
My battery is 39 percent remaining.
I want to go back home to charge my battery.
```

I tested this PR by setting the receiver to my email address

PR #1388, opened 2021-09-21
Add ros node to send email based on received ros topic

708yamaguchi:
From https://github.com/jsk-ros-pkg/jsk_robot/pull/1374#issuecomment-923757369

I add ros node to send email based on received ros topic.
Default email address can be set by using `/var/lib/robot/email_topic.yaml`

Usage:
```bash
$ catkin build jsk_robot_startup
$ rosrun jsk_robot_startup email_topic.py
$ rostopic pub /email jsk_robot_startup/Email "header:
  seq: 0
  stamp: {secs: 0, nsecs: 0}
  frame_id: ''
sender_address: '708yamaguchi@gmail.com'
receiver_address: '708yamaguchi@gmail.com'
subject: 'email test'
body: 'body test'" 
```

### DIFF
--- a/jsk_robot_common/jsk_robot_startup/CMakeLists.txt
+++ b/jsk_robot_common/jsk_robot_startup/CMakeLists.txt
@@ -57,6 +57,7 @@ generate_dynamic_reconfigure_options(
 add_message_files(
   FILES
   RoomLight.msg
+  Email.msg
 )
 
 generate_messages(
--- a/jsk_robot_common/jsk_robot_startup/msg/Email.msg
+++ b/jsk_robot_common/jsk_robot_startup/msg/Email.msg
@@ -1,0 +1,5 @@
+std_msgs/Header header
+string sender_address # Sender's email address
+string receiver_address # receiver's email address
+string subject # Email subject
+string body # Email body
--- a/jsk_robot_common/jsk_robot_startup/scripts/email_topic.py
+++ b/jsk_robot_common/jsk_robot_startup/scripts/email_topic.py
@@ -1,0 +1,58 @@
+#!/usr/bin/env python
+
+# automatically reset slam as robot put on the ground
+
+from jsk_robot_startup.msg import Email
+import os
+import rospy
+import subprocess
+import yaml
+
+
+class EmailTopic(object):
+    def __init__(self):
+        address_yaml = rospy.get_param(
+            '~address_yaml', "/var/lib/robot/email_topic.yaml")
+        if os.path.exists(address_yaml):
+            with open(address_yaml) as yaml_f:
+                yaml_addresses = yaml.load(yaml_f)
+                self.sender_address = yaml_addresses['sender_address']
+                self.receiver_address = yaml_addresses['receiver_address']
+        self.subscriber = rospy.Subscriber(
+            'email', Email, self._cb, queue_size=1)
+
+    def _cb(self, msg):
+        # If sender_address or receiver_address is empty, use address in yaml
+        if msg.sender_address == '':
+            sender_address = self.sender_address
+        else:
+            sender_address = msg.sender_address
+        if msg.receiver_address == '':
+            receiver_address = self.receiver_address
+        else:
+            receiver_address = msg.receiver_address
+        # Send email
+        self._send_mail(
+            msg.subject, msg.body, sender_address, receiver_address)
+
+    def _send_mail(self, subject, body, sender_address, receiver_address):
+        cmd = "LC_CTYPE=en_US.UTF-8 /bin/echo -e \"{}\"".format(body)
+        cmd += " | /usr/bin/mail -s \"{}\" -r {} {}".format(
+            subject, sender_address, receiver_address)
+        exit_code = subprocess.call(cmd, shell=True)
+        rospy.loginfo('Title: {}'.format(subject))
+        if exit_code > 0:
+            rospy.logerr(
+                'Failed to send e-mail:  {} -> {}'.format(
+                    sender_address, receiver_address))
+            rospy.logerr("You may need to do '$ sudo apt install mailutils'")
+        else:
+            rospy.loginfo(
+                'Succeeded to send e-mail: {} -> {}'.format(
+                    sender_address, receiver_address))
+
+
+if __name__ == "__main__":
+    rospy.init_node("email_topic")
+    app = EmailTopic()
+    rospy.spin()


k-okada:
nice,  I think we'd like to send e-mail with image attached. So please add attachments section. To do this, please consider using https://docs.python.org/ja/2.7/library/smtplib.html, This is standard library and seems no big difference between python2/3.
I am afraid it becomes very difficult if we need to handle password, but may be just connecting localhost will be ok. 
https://github.com/WPI-ARC/arclab_ros/blob/master/ros_email/src/ros_email/smtp_email.py

708yamaguchi:
It seems that we can send attached file with `Mailutils` (-A option)
https://manpages.debian.org/testing/mailutils/mail.1.en.html

Anyway, I rewrite this node with `smtplib`.

By the way, I found that we can send email with attached file to external address if we launch smtp server at local.

For example,

launch smtp server with `postfix`:
```
sudo apt install postfix
# sudo systemctl status postfix
# sudo systemctl start postfix
```

send email with `Mailutils`:
```
sudo apt install mailutils
echo "content" | mail -s "title" [receiver_email_address] -A [attached_file_path]
```

send email with `smtplib`: (I don't know the reason, but I was able to send gmail -> jsk address, but not able to send jsk address -> gmail)
```
#!/usr/bin/env python
# -*- coding: utf-8 -*-

import smtplib
import os
from email.mime.multipart import MIMEMultipart
from email.mime.text import MIMEText
from email.mime.application import MIMEApplication

stmp_server = "localhost"
stmp_port = 25

to_address = "[receiver_email_address]"
from_address = "[sender_email_address]"
subject = "Python Test Email"
body = """
<html>
    <body>
        <h1>Eメールの送信テストです。</h1>
        <p>よろしくお願いします。</p>
    </body>
</html>"""

filepath = "[attached_file_path]"
filename = os.path.basename(filepath)

msg = MIMEMultipart()
msg["Subject"] = subject
msg["From"] = from_address
msg["To"] = to_address
msg.attach(MIMEText(body, "html"))

with open(filepath, "rb") as f:
    mb = MIMEApplication(f.read())

mb.add_header("Content-Disposition", "attachment", filename=filename)
msg.attach(mb)

s = smtplib.SMTP(stmp_server, stmp_port)
s.starttls()
s.sendmail(from_address, to_address, msg.as_string())
s.quit()

print("Eメールを送信しました。")
```

k-okada:
thank you, just to make sure

> send email with smtplib: (I don't know the reason, but I was able to send gmail -> jsk address, but not able to send jsk address -> gmail)


do you mean we can send to gmail adderss with `mail` command, but not `smtplib`? 
I could not send email by `echo  "hello world" | mail おかだ@gmail.com`. Can you?

I think this is because of mail server of gmail have strict authentication check, if you look at `/var/log/mail.log` of your machine, you can find something like.
```
Oct  5 18:30:15 p51s postfix/smtp[22327]: BAD431CE3FC9: host gmail-smtp-in.l.google.com[64.233.189.26] said: 421-4.7.0 This message does not have authentication information or fails to pass 421-4.7.0 authentication checks. To best protect our users from spam, the 421-4.7.0 message has been blocked. Please visit 421-4.7.0  https://support.google.com/mail/answer/81126#authentication for more 421 4.7.0 information. y126si22182176pfb.277 - gsmtp (in reply to end of DATA command)
```

Instead, if you set
```
smtp_server: 'mail.server.address'                                         
```
Then, you can send email to google adderss.

To find `'mail.server.address'`, try
```
 dig ジェーエスケーティー．．．． MX
```
So, I think 
```
sender_address: '708yamaguchi@gmail.com'
```
is not so important, and usually we should use default address, to show who actually send e-mail. Thus I send https://github.com/708yamaguchi/jsk_robot/pull/16 PR.

knorth55:
@k-okada 
> I could not send email by echo "hello world" | mail おかだ@gmail.com. Can you?

with this following settings, you can send gmail from `mail` command.
https://kifarunix.com/configure-postfix-to-use-gmail-smtp-on-ubuntu-18-04/

k-okada:
I think it equivalent to set smtp_server,
--
◉ Kei Okada



2021年10月5日(火) 18:43 Shingo Kitagawa ***@***.***>:

708yamaguchi:
@k-okada 

> do you mean we can send to gmail adderss with mail command, but not smtplib?
I could not send email by echo "hello world" | mail おかだ@gmail.com. Can you?

Yes, I can send email by `echo "hello world" | mail 708yamaguchi@gmail.com`.
Though all of the email are regarded as spam, the authentication error does not occur.

However, when I use python smtplib with localhost smtp server, I got the authentication error.
```
Oct  5 21:45:52 b4apc postfix/smtp[13504]: 9B8C915E31B8: host gmail-smtp-in.l.google.com[108.177.125.26] said: 421-4.7.0 This message does not have authentication information or fails to pass 421-4.7.0 authentication checks. To best protect our users from spam, the 421-4.7.0 message has been blocked. Please visit 421-4.7.0  https://support.google.com/mail/answer/81126#authentication for more 421 4.7.0 information. u2si1849439pjd.165 - gsmtp (in reply to end of DATA command)
Oct  5 21:45:55 b4apc postfix/smtp[13504]: 9B8C915E31B8: to=<708yamaguchi@gmail.com>, relay=alt1.gmail-smtp-in.l.google.com[142.250.141.27]:25, delay=5.5, delays=0.01/0/4.2/1.4, dsn=4.7.0, status=deferred (host alt1.gmail-smtp-in.l.google.com[142.250.141.27] said: 421-4.7.0 This message does not have authentication information or fails to pass 421-4.7.0 authentication checks. To best protect our users from spam, the 421-4.7.0 message has been blocked. Please visit 421-4.7.0  https://support.google.com/mail/answer/81126#authentication for more 421 4.7.0 information. pg7si2191650pjb.45 - gsmtp (in reply to end of DATA command))
```

When I use python smtplib with jsk smtp server, I got the folowing python error.
```
Traceback (most recent call last):
  File "test.py", line 26, in <module>
    server.sendmail(from_email, to_email, msg.as_string())
  File "/usr/lib/python2.7/smtplib.py", line 748, in sendmail
    raise SMTPRecipientsRefused(senderrs)
smtplib.SMTPRecipientsRefused: {'708yamaguchi@gmail.com': (554, '5.7.1 <708yamaguchi@gmail.com>: Relay access denied')}
```

k-okada:
> Yes, I can send email by echo "hello world" | mail 708yamaguchi@gmail.com.

what is the `relayhost` section of `/etc/postfix/main.cf` ?

708yamaguchi:
My relayhost is as follows:
```
$ cat /etc/postfix/main.cf | grep relayhost
relayhost = 
```

708yamaguchi:
@k-okada 

I can send email to gmail with both `mail` and `smtplib`, too.

In the previous comment, I said `I don't know the reason, but I was able to send gmail -> jsk address, but not able to send jsk address -> gmail`, but I find this is my mistake. (I'm so sorry...)

I edited `/var/lib/robot/email_info.yaml` file but I did not care about it.
After I removed `/var/lib/robot/email_info.yaml` file, I can send email to gmail.

My postfix setting:
```
$ cat /etc/postfix/main.cf | grep relay
smtpd_relay_restrictions = permit_mynetworks permit_sasl_authenticated defer_unauth_destination
relayhost = 
```

email script:
```
$ rostopic pub /email jsk_robot_startup/Email "header:
  seq: 0
  stamp: {secs: 0, nsecs: 0}
  frame_id: ''
sender_address: ''
receiver_address: '708yamaguchi@gmail.com'
subject: 'email'
body: 'body test'"
```

mail log
```
Oct  8 15:18:13 b4apc postfix/smtp[14212]: 7F22B15E31B3: to=<708yamaguchi@gmail.com>, relay=gmail-smtp-in.l.google.com[74.125.203.26]:25, delay=2.8, delays=0.01/0/1.3/1.5, dsn=2.0.0, status=sent (250 2.0.0 OK  1633673893 h14si25080960pfh.218 - gsmtp)
Oct  8 15:18:13 b4apc postfix/qmgr[12293]: 7F22B15E31B3: removed
```

k-okada:
@708yamaguchi BTW, can you restart GithubAction by yourself?

708yamaguchi:
> BTW, can you restart GithubAction by yourself?

Yes, I can.

PR #1395, opened 2021-09-28
Add :bodies check to fix travis failure

708yamaguchi:
I hope this PR fix travis failure.

https://github.com/jsk-ros-pkg/jsk_robot/pull/1213#issuecomment-929304356

### DIFF
--- a/.travis.rosinstall
+++ b/.travis.rosinstall
@@ -3,10 +3,12 @@
 # waiting for release 0.3.15
 # we need :go-pos-unsafe bug fix commit for fetch navigation test
 # see https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/425
+# we need :bodies check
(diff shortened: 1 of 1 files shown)


k-okada:
thank you, can we use
```
- git:
    local-name: jsk-ros-pkg/jsk_pr2eus
    uri: https://github.com/jsk-ros-pkg/jsk_pr2eus.git
    version: master
```
or df0fc28a965168744c7d2bd1dd57e6fca8fb9d29 ?

708yamaguchi:
I think we can use both `df0fc28a965168744c7d2bd1dd57e6fca8fb9d29` and `master`.

Should we use these latest commit?

k-okada:
I think if we can the latest commit on CI, that would be very helpful.
--
◉ Kei Okada



2021年9月29日(水) 16:38 Naoya Yamaguchi ***@***.***>:

PR #1409, opened 2021-11-11
Update email topic node

sktometometo:
- Support multiple attachments
- Add client library, example and demo launch
- Update README

PR #1426, opened 2021-12-25
Add node to shut down or reboot robot itself

708yamaguchi:
Add node to shutdown/reboot robot by rostopic.

This node can be used, for example, to shut down the robot so that it does not consume battery power when it gets navigation stuck.

Note that this node needs to be started with sudo privileges.

### DIFF
--- a/jsk_robot_common/jsk_robot_startup/scripts/shutdown.py
+++ b/jsk_robot_common/jsk_robot_startup/scripts/shutdown.py
@@ -1,0 +1,50 @@
+#!/usr/bin/env python
+
+import os
+import rospy
+from std_msgs.msg import Empty
+
+
+class Shutdown(object):
+    """
+    This node shuts down or reboots the robot itself
+    according to the rostopic.
+
+    Note that this node needs to be run with sudo privileges.
+
+    Usage:
+    # Launch node
+    $ su [sudo user] -c ". [setup.bash]; rosrun jsk_robot_startup shutdown.py"
+
+    # To shutdown robot
+    rostopic pub /shutdown std_msgs/Empty
+    # To restart robot
+    rostopic pub /reboot std_msgs/Empty
+    """
+
+    def __init__(self):
+        rospy.loginfo('Start shutdown node.')
+        rospy.Subscriber('shutdown', Empty, self.shutdown)
+        rospy.Subscriber('reboot', Empty, self.reboot)
+
+    def shutdown(self, msg):
+        rospy.loginfo('Shut down robot.')
+        shutdown_command = '/sbin/shutdown -h now'
+        ret = os.system(shutdown_command)
+        if ret != 0:
+            rospy.logerr("Failed to call '$ {}'. Check authentication.".format(
+                shutdown_command))
+
+    def reboot(self, msg):
+        rospy.loginfo('Reboot robot.')
+        reboot_command = '/sbin/shutdown -r now'
+        ret = os.system(reboot_command)
+        if ret != 0:
+            rospy.logerr("Failed to call '$ {}'. Check authentication.".format(
+                reboot_command))
+
+
+if __name__ == '__main__':
+    rospy.init_node('shutdown')
+    s = Shutdown()
+    rospy.spin()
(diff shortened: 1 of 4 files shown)


jsk_robot_common/jsk_robot_startup/scripts/shutdown.py
```diff
@@ -0,0 +1,50 @@
...
+    """
+
+    def __init__(self):
+        rospy.loginfo('Start shutdown node.')
+        rospy.Subscriber('shutdown', Empty, self.shutdown)
+        rospy.Subscriber('reboot', Empty, self.reboot)
+
+    def shutdown(self, msg):
+        rospy.loginfo('Shut down robot.')
+        shutdown_command = '/sbin/shutdown -h now'
```
knorth55, line 32: can you change this node to customize shutdown command from `rosparam`?
for pr2, we need to use `pr2-shutdown`

jsk_robot_common/jsk_robot_startup/scripts/shutdown.py
```diff
@@ -0,0 +1,50 @@
...
+        ret = os.system(shutdown_command)
+        if ret != 0:
+            rospy.logerr("Failed to call '$ {}'. Check authentication.".format(
+                shutdown_command))
+
+    def reboot(self, msg):
+        rospy.loginfo('Reboot robot.')
+        reboot_command = '/sbin/shutdown -r now'
```
knorth55, line 40: the same comment as `shutdown`.
I want to customize the command with `rosparam`

knorth55 (changes requested):

jsk_robot_common/jsk_robot_startup/scripts/shutdown.py
708yamaguchi: Thank you for your review!
That makes sense.

I updated.
[`661b4f6` (#1426)](https://github.com/jsk-ros-pkg/jsk_robot/pull/1426/commits/661b4f6a42dd32b39e0b4e3841c229be93b53f48)

jsk_robot_common/jsk_robot_startup/scripts/shutdown.py
708yamaguchi: Thank you.
I updated the same as shutdown.

knorth55 (approved):

k-okada:
@708yamaguchi can we publish topic to https://github.com/jsk-ros-pkg/jsk_robot/pull/1409, just before `shutdown` ?, so that we can know if the system calls shutdown command

k-okada:
@708yamaguchi May be publish topic to send email is not `shutdown.py`'s work.  
The power?battery? management node? should send `shutdown` topic after sending email. Please add PR to such nodes.

708yamaguchi:
@k-okada 

Thank you very much for your review.

I added `power_management.py.
This node
1. subscribes `~shutdown` and `~reboot` topic
2. send email by publishing `/email
3. shutdown/reboot robot by publishing `/shutdown` and `/reboot`

`power_management` Document:
https://github.com/708yamaguchi/jsk_robot/blob/shutdown-node-master/jsk_robot_common/jsk_robot_startup/README.md#scriptspower_managementpy

k-okada:
I think fetch robot has script to read battery status and publish `shoutdown` topic.
And my guess is; since each robot have different battery system, each robot need to have something like power_management.py, So what we need is the list of power_management.py in the document..

708yamaguchi:
@k-okada 

> So what we need is the list of power_management.py in the document..

I'm sorry, but I may not understand the meaning of this very well.

In my understanding, you mean I should add document like the following to `jsk_pr2_startup`, `jsk_fetch_startup` README.

For example, jsk_fetch_startup README,
```markdown
## Power management nodes

### jsk_robot_startup/power_management.py

This node shutdown/reboot robot.

### jsk_fetch_startup/battery_warning.py

This node notifies users of low battery by email.
```

k-okada:
jsk_robot_startup contains common nodes for (at least jsk's) robot, so if
you put power_management, you need at least more than one robot that uses
this system.
If you have only one example, I guess power_management.py is only designed
for fetch robot, we should put that function within jsk_fetch_startup, in
that case, we could update battery_warning.py to publish 'shutdown' in
addition to https://github.com/jsk-ros-pkg/jsk_robot/pull/1374/files.
If you'd like to put power_managemnt within jsk_robot_startup, that is a
nice thing. But In that case, please consider how can you put that effort
within your thesis.
--
◉ Kei Okada



2021年12月27日(月) 23:22 Naoya Yamaguchi ***@***.***>:

708yamaguchi:
@k-okada 

> we could update battery_warning.py to publish 'shutdown' in addition to #1374 (files).

Thank you very much for your advice.
Move to https://github.com/jsk-ros-pkg/jsk_robot/pull/1431

PR #1434, opened 2022-01-17
Remove empty 'kinova' package

Affonso-Gui:
kinova-interface.l has an unnecessary package definition for "kinova" (also, correct (usual) syntax would be with capital letters).

@a-ichikura


Affonso-Gui:
Moved to https://github.com/sktometometo/jsk_robot/pull/134

PR #1465, opened 2022-04-20
Add supervisor job for edgetpu-panorama recognition

708yamaguchi:
Please merge after https://github.com/jsk-ros-pkg/jsk_robot/pull/1464

I cherry-pick
- https://github.com/knorth55/jsk_robot/commit/ef53115f5b99725f3566df4d63b510c33fc84e49
- https://github.com/knorth55/jsk_robot/commit/1ac12cbddd8b2eb29b7cb9921153d867972cb8da
- https://github.com/knorth55/jsk_robot/pull/112

### DIFF
--- a/jsk_fetch_robot/jsk_fetch_startup/apps/go_to_kitchen/go_to_kitchen.app
+++ b/jsk_fetch_robot/jsk_fetch_startup/apps/go_to_kitchen/go_to_kitchen.app
@@ -26,6 +26,13 @@ plugins:
       video_title: go_to_kitchen_object_detection.avi
       video_topic_name: /edgetpu_object_detector_visualization/output
       video_fps: 5.0
+  - name: panorama_video_recorder_plugin
+    type: app_recorder/video_recorder_plugin
+    launch_args:
+      video_path: /tmp
+      video_title: go_to_kitchen_panorama.avi
+      video_topic_name: /dual_fisheye_to_panorama/output
+      video_fps: 1.0
   - name: respeaker_audio_recorder_plugin
     type: app_recorder/audio_recorder_plugin
     launch_args:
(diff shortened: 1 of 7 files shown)


knorth55 (approved):

k-okada:
please discuss with @708yamaguchi @sktometometo ,
Since auto starting function is so necessary and useful, I think it is time to establish common tools/method to put them in our  robots,

My comment is
1) Sometimes it is better to investigate different ideas, like tendon robot vs geared robot, lider vs vision, but I think we have no reason to keep systemd vs supervisor 
2) Since systemd/supervisor is a robot agnostic tool, we could put the main code in jsk_robot_comon, like start roscore, app_manager, ...
3) It always nice to imagine someone is doing same thing, I haven't use https://github.com/magazino/systemd_ros by myself, but I think they use this on their product and they always have talented engineers, so there are good reason to try them.

BTW, What system does Stratch use ?

knorth55:
IMO, `systemd` might be better with `Cockpit` than `supervisor`, but the motivation is just cool GUI.
So I'm not super excited to switch from `supervisor` to `systemd`.
(Also I don't trust `systemd`, it will change soon as `init.d`, `service`, `upstart`...)

> Since systemd/supervisor is a robot agnostic tool, we could put the main code in jsk_robot_comon, like start roscore, app_manager, ...

This idea sounds good, but I don't have a good idea.
Because JSK robots are targeted to start all nodes with one or two launch, `systemd` config may be just one or two.
Other non-ROS services can be the same for all robots, so I can make the `jsk_robot_common/jsk_robot_systemd`, but I'm not so motivated...

> It always nice to imagine someone is doing same thing, I haven't use https://github.com/magazino/systemd_ros by myself, but I think they use this on their product and they always have talented engineers, so there are good reason to try them.

I know `systemd_ros`, but hmmm.
If all robots  uses `systemd` and generate the `systemd` conf in `jsk_robot`, that would be great.
But the reason why I'm not motivated is that `systemd` cannot be on `PR2` + `Indigo`.
We can make the system for `Fetch` and `Spot`, but `Spot` `systemd` is well-handcrafted and super complicated...

knorth55:
`Stretch` uses default `Startup Applications` in `Ubuntu`.
https://help.ubuntu.com/stable/ubuntu-help/startup-applications.html.en

k-okada:
> Because JSK robots are targeted to start all nodes with one or two launch, systemd config may be just one or two, so we can write down manually.

That's good point, I also prefer using  jsk_<robot>_startup.launch for systemd/supervisor, and if we do so, I think using different system for different robot may not a big problem.
For example, this PR essentially add one node to one launch file, but require to update 7 files and ask other people to review.

But, If we need to run each functions with different terminal using tmux, then we need smart way to keep our configuration and develop process more simple.

> Stretch uses default Startup Applications in Ubuntu.

They are smart, as always...

sktometometo:
@708yamaguchi Could you resolve conflicts?

PR #1485, opened 2022-05-24
[jsk_robot_startup] Support embed image in email_topic.py

tkmtnt7000:
This PR supports sending embed image with email_topic.py. (Discussion in #1463)

------
TODO
- [x] Set image size 
---------

- sample code
```python
#!/usr/bin/env python
# -*- coding: utf-8 -*-

import rospy
import time
from jsk_robot_startup.msg import Email
from jsk_robot_startup.msg import EmailBody

def main():
    rospy.init_node('embed_image_test')
    pub = rospy.Publisher("email", Email, queue_size=10)

    msg = Email()
    content_1 = EmailBody()
    content_1.type = 'text'
    content_1.message = 'こんにちは．\n'
    content_2 = EmailBody()
    content_2.type = 'img'
    content_2.file_path = '/home/tsukamoto/Pictures/spotkinova_attachment.jpg'
    content_4 = EmailBody()
    content_4.type = 'img'
    content_4.file_path = '/home/tsukamoto/Pictures/boston_template.png'
    content_5 = EmailBody()
    content_5.type = 'text'
    content_5.message = '\n 今日はSpotで遊んだよ\n'
    content_6 = EmailBody()
    content_6.type = 'img'
    content_6.file_path = '/home/tsukamoto/Pictures/fetch-door4.png'
    content_7 = EmailBody()
    content_7.type = 'text'
    content_7.message = '\n Fetchにも手伝ってもらったよ．\n'
    content_8 = EmailBody()
    content_8.type = 'text'
    content_8.message = '\n'

    msg.body = [content_1, content_2, content_8, content_4, content_5, content_6, content_7]
    msg.subject = 'embed_test'
    msg.sender_address = 'hoge@hoge.com'
    msg.receiver_address = 'fuga@fuga.com'
    print (msg)

    time.sleep(1)
    rospy.loginfo("Publish")
    pub.publish(msg)

main()

```

### DIFF
--- a/jsk_robot_common/jsk_robot_startup/CMakeLists.txt
+++ b/jsk_robot_common/jsk_robot_startup/CMakeLists.txt
@@ -59,6 +59,7 @@ add_message_files(
   FILES
   RoomLight.msg
   Email.msg
+  EmailBody.msg
 )
 
 generate_messages(
--- a/jsk_robot_common/jsk_robot_startup/euslisp/email-topic-client.l
+++ b/jsk_robot_common/jsk_robot_startup/euslisp/email-topic-client.l
@@ -1,5 +1,7 @@
 (ros::load-ros-manifest "jsk_robot_startup")
 
+(require :base64 "lib/llib/base64.l")
+
 (defun init-mail ()
   (ros::advertise "email" jsk_robot_startup::Email 1)
   (ros::spin-once)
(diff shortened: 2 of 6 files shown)


knorth55:
Superb!
In example code, where do you add the first and second image?

tkmtnt7000:
> In example code, where do you add the first and second image?

Sorry, I mistakenly write an example code. Updated.

708yamaguchi:
Thank you for this cool feature.

I also checked that `html` type worked well.

```python
    content_1 = EmailBody()
    content_1.type = 'html'
    content_1.message = '<p><font size="7" color="#00ff00">サイズが7で緑色</font></p>\n'
```

One thing.
Do we need to use `\n` like below?
If possible, I do not want to write `\n` in the `message` field.
```
    content_5.message = '\n 今日はSpotで遊んだよ\n'
```

708yamaguchi:
> One thing.
Do we need to use `\n` like below?
If possible, I do not want to write `\n` in the `message` field. (Maybe my original code is bad, sorry)
```
    content_5.message = '\n 今日はSpotで遊んだよ\n'
```

Sorry, `\n` is needed when we want to write multiple line message.

tkmtnt7000:
Thank you for comments.
The size of embedded images can now be specified as a percentage.

Now, we have to write `\n` manually as a text type when multiple images are to be aligned vetically.
It is possible to automatically break each image into a new line, but I feel that there is a demand to arrange images horizontally...

knorth55:
the release is a good point of jsk_common, hmm.
currently if jaxon wants to send e-mail, they need to install jsk_robot, which is a bit unfriendly.
I'm for to move this node to jsk_common.

tkmtnt7000:
So...   should I move `email_topic.py` to https://github.com/jsk-ros-pkg/jsk_common ???

knorth55:
> But Jaxon robot should depend on jsk_robot_startup and uses life log to store sensor/event data and use them to learn something from experiments….

I agree.
I think both idea is acceptable, so its better to leave it in jsk_robot now.

tkmtnt7000:
Sorry, I understood mistakenly.
I leave this node in jsk_robot.

Affonso-Gui (commented):
Arguably it would also be nice to have an option to send an image directly from the client, instead of needing a local path in the robot PC. (I and Kochigami are working on a similar case for Pepper's tablet,  thinking of further usage with scratch)

The message could be encoded as base64 or any other ros-compliant data type.
e.g.
```
[app_manager/Icon]:
string format
uint8[] data
```

We should evaluate if this is safe or not first, though.

tkmtnt7000:
When I was writing a node that sends an email with an image depending on the state of smach, I realized that it would be nice to receive the image topic(`sensor_msgs/Image`) directly.
https://github.com/k-okada/jsk_robot/pull/60

PR #1487, opened 2022-05-25
[jsk_fetch_startup] Use mongodb_store which can manage logerr throttle rate

708yamaguchi:
From https://github.com/knorth55/jsk_robot/pull/237

When using fetch, it is difficult to debug because it sometimes throws out a large number of errors such as the following.
When I parsed log file, 250000 lines error is output for 10000 seconds.

```
[ERROR] [1653231340.987602] [/replicator_node:rosout]: [mongorestore] - E11000 duplicate key error collection: jsk_robot_lifelog.fetch1075 index: _id_ dup key: { : ObjectId('6243af9651998d10f0c7787c') }
```

By changing the code in `mongodb_store`, I throttle the error log.
https://github.com/strands-project/mongodb_store/pull/271


k-okada (approved):
merge ok, please pass CI

PR #1497, opened 2022-06-03
[jsk_pr2_startup] Enable additional topic and options in rosbag_record.launch

nakane11:
Other than topics specified by default, you can record any topics including regex. 
You can also add other options as [common_record.launch](https://github.com/jsk-ros-pkg/jsk_common/blob/master/jsk_data/launch/common_record.launch) in jsk_data.

cc: @iory


YUKINA-3252:
The following command was used to record rosbag.
```
roslaunch jsk_pr2_startup rosbag_record.launch other_topics:=/edgetpu_object_detector_petal/output/image/compressed rosbag:=/home/tork/.ros/2022-06-06-18-38.bag
```
I was able to check `/edgetpu_object_detector_petal/output/image/compressed' with the following command.
```
roslaunch jsk_pr2_startup rosbag_play.launch rosbag:=/home/tork/.ros/2022-06-06-18-38.bag gui:=true
```

708yamaguchi:
@YUKINA-3252 

Nice try!

PR #1499, opened 2022-06-03
[jsk_fetch_startup] Do not use ConnectionBasedTransport in correct_position.py

708yamaguchi:
From https://github.com/knorth55/jsk_robot/pull/254

Since https://github.com/jsk-ros-pkg/jsk_common/pull/1711 is merged, `correct_position.py` output the following diagnostics error.

```
"/Other/correct_position:  correct_position"
Other correct position:  correct position  correct position is not running for 1.0 sec
```

This is because `correct_position.py` does not publish `/initialpose` (in other words `self.last_published_time` is not updated) when fetch is un-docked.

With this PR, I suppress the diagnostics error by removing `ConnectionBasedTransport` from `correct_position.py`.
I think this PR does not increase fetch's CPU usage very much, because `correct_position.py` does not use CPU resource even if `/initialpose` is always published .

### DIFF
--- a/jsk_fetch_robot/jsk_fetch_startup/scripts/correct_position.py
+++ b/jsk_fetch_robot/jsk_fetch_startup/scripts/correct_position.py
@@ -2,20 +2,20 @@
 
 import rospy
 
-from jsk_topic_tools import ConnectionBasedTransport
 from power_msgs.msg import BatteryState
 from geometry_msgs.msg import PoseWithCovarianceStamped
 from geometry_msgs.msg import Pose
 from visualization_msgs.msg import MarkerArray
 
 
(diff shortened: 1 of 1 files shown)


iory (approved):

knorth55:
I don't think it is a good idea to change the existing node and adopt to new feature.
This PR is needed because `Timer` publisher is not considered in the `ConnectionBasedTransport`, right.
Hmmm, it becomes more difficult to write node which works properly...

708yamaguchi:
Thank you for your advice.

Based on offline talk and https://github.com/jsk-ros-pkg/jsk_common/pull/1740, move to https://github.com/jsk-ros-pkg/jsk_robot/pull/1500

PR #1517, opened 2022-06-30
[jsk_robot_startup/util] Support python3 in mux_selector.py

tkmtnt7000:
In python3, we have to use `_thread` instead of `thread`.

### DIFF
--- a/jsk_robot_common/jsk_robot_startup/util/mux_selector.py
+++ b/jsk_robot_common/jsk_robot_startup/util/mux_selector.py
@@ -24,10 +24,11 @@ import roslib.message
 import rostopic
 from topic_tools.srv import MuxSelect
 import traceback
-if sys.version.major == 2:
+if sys.version_info.major < 3:
     import thread
 else:
-    import _thread
+    import _thread as thread
+
 
 def callback (m, expr, topic, index):
     global selects, lockobj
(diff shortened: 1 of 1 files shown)


jsk_robot_common/jsk_robot_startup/util/mux_selector.py
```diff
@@ -19,13 +19,15 @@
...
 import sys
 import roslib.message
 import rostopic
 from topic_tools.srv import MuxSelect
 import traceback
-
+if sys.version.major == 2:
+    import thread
+else:
+    import _thread
```
iory, line 30: How about writing `import _thread` as `import _thread as thread`?

iory (changes requested):

jsk_robot_common/jsk_robot_startup/util/mux_selector.py
tkmtnt7000: Thank you for advice. I think that's a good idea.
Modified in [`642da39` (#1517)](https://github.com/jsk-ros-pkg/jsk_robot/pull/1517/commits/642da39f4a9f6e8298ced5e6cbbcfca0d9c7825f)

tkmtnt7000: sorry, I made a mistake.

iory: Following changes will be good.
```
import sys
if sys.version_info.major < 3:
    import thread
else:
    import _thread as thread
```

tkmtnt7000: Thanks. I fixed.

iory (approved):

PR #1401, opened 2021-10-09
Add script to bind rfcomm device to remote Bluetooth device

708yamaguchi:
cherry-picked from https://github.com/knorth55/jsk_robot/pull/146

I add script to bind rfcomm device to remote bluetooth device.

By binding rfcomm device, we can connect bluetooth device via rosserial.
https://github.com/sktometometo/M5Stack_Bluetooth_rosserial_example

#### Usage

Save the bluetooth device MAC address to file like (`/var/lib/robot/rfcomm_devices.yaml`)

```
- name: device1
  address: XX:XX:XX:XX:XX:XX
- name: device2
  address: YY:YY:YY:YY:YY:YY
```

Then, run this program.
If you use PR2,
```
roslaunch jsk_pr2_startup pr2_rfcomm_bind.launch machine:=c2  # bluetooth dongle is now in c2
```

#### management

Currently, I automatically start the above launch file by upstart in PR2.

The config file is in /etc/upstart/jsk-rfcomm-bind.conf in pr2.

To check how many devices are bound to rfcomm, use rfcomm command.

```
ssh pr2
ssh c2
rfcomm
```

PR #1519, opened 2022-06-30
[jsk_fetch_startup] Add launch to use m5stack sensors via rosserial + bluetooth

708yamaguchi:
Please merge after https://github.com/jsk-ros-pkg/jsk_robot/pull/1401

With this PR, fetch1075 can always communicate with m5stacks, which is usually powered by a refrigerator.
- When m5stack is charged, fetch1075 can get the m5stack battery rostopics. (The sensors are not available to save m5stack battery)
- When m5stack is not charged, fetch1075 can get each sensor rostopics in addition to the battery rostopics.

Only fetch1075 uses m5stacks because two robots (fetch15 and fetch1075) cannot use one rosserial device at the same time.

m5stacks_charged

### DIFF
--- a/jsk_fetch_robot/jsk_fetch_startup/launch/fetch_bringup.launch
+++ b/jsk_fetch_robot/jsk_fetch_startup/launch/fetch_bringup.launch
@@ -259,4 +259,9 @@
     </rosparam>
   </node>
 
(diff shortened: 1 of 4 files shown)


708yamaguchi:
From https://github.com/knorth55/jsk_robot/pull/278

PR #1525, opened 2022-07-06
Erase the last element of the vector of reset-pose and correct vector…

YUKINA-3252:
This PR fixes the length of the vector in 'reset-pose'  from 15 to 16 by erasing the last element of the vector.

This was found by the following warning issued https://github.com/euslisp/jskeus/pull/612#issue-1118454435.
```
;; WARNING : Dimension of input vector: 16 > Dimension of joint-list: 15
```


knorth55:
so i mean this PR is correct!
baxter's knod joint should be ignored snd total num of angle-vector should be 15.

PR #1441, opened 2022-03-31
initial unitree package

k-okada:
go1

see attached gist for setup workspace
https://gist.github.com/k-okada/a76e4e678c73683b9ac26c7945daa381

PR #1528, opened 2022-07-10
add cross-compile environment for pepper robot

k-okada:
This branch contains tools/configuration files for cross-compile environment of Pepper robot, by using this you can run your application within Pepper's environment PC
Scripts are mostoriy copied from https://github.com/jsk-ros-pkg/jsk_robot/pull/1441, if we have another robot, which requires cross-compile environment, we should create jsk_robot_cross package


jsk_naoqi_robot/jsk_pepper_startup/apps/meeting/scripts/meeting.l
```diff
@@ -0,0 +1,194 @@
...
+    (send *ri* :speak "\\vct=120\\\\rspd=80\\おつかれさまでしたッ！\\rspd=100\\")
+    (send *ri* :wait-interpolation)
+    (send *ri* :hide-image)
+    (send *ri* :set-language "English")
+    ;;
+    (set-alist 'description "End Meeting" args)
+    :finished))
+
+(defun comment-func (args)
+  (let (count str (wait 600) tm)
```
kochigami, line 131: In `jsk_pepper_startup/apps/meeting/scripts/meeting.l`,
I found the wait time was 60 sec, but right time should be 600 sec.
If the wait time is 60 sec, Pepper speaks too many times.

I modified this, but I mistakenly edited it directly without sending a pull-request... Sorry for that.

k-okada: @kochigami thank you!

jsk_naoqi_robot/jsk_pepper_startup/apps/meeting/scripts/meeting.l
```diff
@@ -0,0 +1,204 @@
...
+    sm))
+
+;; create robot interface
+(unless (boundp '*ri*) (pepper-init))
+(objects (list *pepper*))
+
+;;
+(defun image_cb (msg)
+  (setq *image* msg))
+(ros::subscribe "/pepper_robot/camera/front/image_raw" sensor_msgs::Image #'image_cb)
```
tkmtnt7000, line 200: I think it might be better to use compressd image if compressd image is published. Or manually compress it before base64encoding.
In my local environment with usb_cam_node(size: 640x480), I got an error about something like too long string and euslisp crashed.
```lisp
...
/opt/ros/noetic/share/euslisp/jskeus/eus/Linux64/bin/irteusgl: ERROR th=0 too long string ;; Segment\
ation Fault.
;; in (write-byte (bin2base64 x) s)
;; You are still in a signal handler.
;;Try reset or throw to upper level as soon as possible.
;; code=1740538864 x=67be82c0 addr=10000
Fatal:
```

k-okada: in that case, how about subscribe compressed image  using
```
-(ros::subscribe "/pepper_robot/camera/front/image_raw" sensor_msgs::Image #'image_cb)
+(ros::subscribe "/pepper_robot/camera/front/image_rect_color/compressed" sensor_msgs::CompressedImage #'image_cb)
```
and receives like-> https://gist.github.com/k-okada/4d0dabe396a6af37ca0511d40b4e21e4#file-smach_to_mail-py-compressed-version

tkmtnt7000: >in that case, how about subscribe compressed image using
 and receives like-> [gist.github.com/k-okada/4d0dabe396a6af37ca0511d40b4e21e4#file-smach_to_mail-py-compressed-version](https://gist.github.com/k-okada/4d0dabe396a6af37ca0511d40b4e21e4#file-smach_to_mail-py-compressed-version)


Yes, I have the almost same solution about compressed version.

PR #1529, opened 2022-07-11
[jsk_fetch_startup] Add alternate rwt_app_chooser URL

708yamaguchi:
Currently, we cannot see app icons on http://tork-a.github.io/visualization_rwt/rwt_app_chooser.
This is because recent browsers (e.g. chrome, firefox) have become more secure.

I added the alternate URL to REDAME.

### DIFF
--- a/jsk_fetch_robot/jsk_fetch_startup/README.md
+++ b/jsk_fetch_robot/jsk_fetch_startup/README.md
@@ -234,7 +234,7 @@ tmuxinator log
 You can not run this on Firefox. Please use Google Chrome.
 
 ### Add fetch to rwt_app_chooser
-1. Access [http://tork-a.github.io/visualization_rwt/rwt_app_chooser](http://tork-a.github.io/visualization_rwt/rwt_app_chooser "website"). Note that this website is not available now. Access `http://fetch's IP address:8000/rwt_app_chooser/`.
+1. Access [http://tork-a.github.io/visualization_rwt/rwt_app_chooser](http://tork-a.github.io/visualization_rwt/rwt_app_chooser "website").
 1. Click `ADD A ROBOT` button
 1. Select `Fetch` at `Robot type`
 1. Type `fetch15` at `Robot name`
--- a/jsk_fetch_robot/jsk_fetch_startup/README.md
+++ b/jsk_fetch_robot/jsk_fetch_startup/README.md
@@ -234,7 +234,7 @@ tmuxinator log
 You can not run this on Firefox. Please use Google Chrome.
 
 ### Add fetch to rwt_app_chooser
-1. Access [http://tork-a.github.io/visualization_rwt/rwt_app_chooser](http://tork-a.github.io/visualization_rwt/rwt_app_chooser "website").
+1. Access [http://tork-a.github.io/visualization_rwt/rwt_app_chooser](http://tork-a.github.io/visualization_rwt/rwt_app_chooser "website"). Note that this website is not available now. Access `http://fetch's IP address:8000/rwt_app_chooser/`.
 1. Click `ADD A ROBOT` button
 1. Select `Fetch` at `Robot type`
 1. Type `fetch15` at `Robot name`
--- a/jsk_fetch_robot/jsk_fetch_startup/README.md
+++ b/jsk_fetch_robot/jsk_fetch_startup/README.md
@@ -235,6 +235,8 @@ You can not run this on Firefox. Please use Google Chrome.
 
 ### Add fetch to rwt_app_chooser
 1. Access [http://tork-a.github.io/visualization_rwt/rwt_app_chooser](http://tork-a.github.io/visualization_rwt/rwt_app_chooser "website").
+  - Be careful to access the site via http, not https, to to enable websocket communication.
+  - Modern browsers may automatically redirect from http to https.
 1. Click `ADD A ROBOT` button
 1. Select `Fetch` at `Robot type`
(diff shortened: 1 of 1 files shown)


knorth55:
with my old chrome, I can see icons.
my chrome version is  100.0.4896.127（Official Build） （64 ビット）

knorth55:
this maybe caused by websocket + SSL security update.
But I didn't check deeply.

708yamaguchi:
Firefox 88.0 can also see the app icons.

k-okada (commented):
What is the root couase? because it hosted on github.io, so all github.io have same problem? or is theare anythig we should fix on http://tork-a.github.io/visualization_rwt ?

708yamaguchi:
I accessed http://tork-a.github.io/visualization_rwt fetch1075 app in my chrome browser.
```
$ google-chrome --version
Google Chrome 100.0.4896.88 
```

However, the app icon did not appear and the following error log was displayed in the browser console.
```
roslib.js:652 Mixed Content: The page at 'https://tork-a.github.io/visualization_rwt/rwt_app_chooser/#!task/fetch1075' was loaded over HTTPS, but attempted to connect to the insecure WebSocket endpoint 'ws://fetch1075.jsk.imi.i.u-tokyo.ac.jp:9090/'. This request has been blocked; this endpoint must be available over WSS.
```
```
roslib.js:652 Uncaught DOMException: Failed to construct 'WebSocket': An insecure WebSocket connection may not be initiated from a page loaded over HTTPS.
    at Ros.connect (https://tork-a.github.io/visualization_rwt/rwt_app_chooser/3rdparty/ros/js/roslib.js:652:24)
    at connect_ros (<anonymous>:180:8)
    at page.onReady (<anonymous>:196:4)
    at https://tork-a.github.io/visualization_rwt/rwt_app_chooser/3rdparty/phonon/js/phonon.js:2492:15
```

This seems to be an error caused by ws:// communication over https://. (So I think it is not a github.io issue.)
The ideal solution I think is to introduce wss:// communication instead of ws:// communication.

Related(?) stackoverflow:
> This is a form of [active mixed content](https://developers.google.com/web/fundamentals/security/prevent-mixed-content/what-is-mixed-content#active_mixed_content). All recent browsers, [such as Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=662692) and Google Chrome prohibit this kind of content, much like they prohibit a secure page (HTTPS loaded page) from loading insecure JavaScript.

> The "correct" way to fix this is to force all 3rd party to HTTPS on their websockets as well and avoid the whole issue.

https://stackoverflow.com/a/51274835

k-okada:
@708yamaguchi thank you , do you mean github.io have strict checking on `ws://` vs `wss://` and fetch's web server does not have them? If so, this is the problem of roswww, rosbrige_server ?
for example
https://github.com/RobotWebTools/rosbridge_suite/issues/232
https://github.com/RobotWebTools/rosbridge_suite/pull/27

708yamaguchi:
> do you mean github.io have strict checking on ws:// vs wss:// and fetch's web server does not have them?

After re-investigating, I think that http:// (not https://) access has not problem.
I check several `rwt_app_chooser` and access `ws://fetch1075.jsk.imi...` with incognito chrome (with no cache mode)

I can access fetch1075's apps:
- `http://localhost:8000/rwt_app_chooser/` (with locally `roslaunch rwt_app_chooser sample.launch`)
- `http://fetch15's IP address:8000/rwt_app_chooser/`
- `http://fetch1075's IP address:8000/rwt_app_chooser/`
- `http://furushchev.ru/rwt_app_chooser/`
- `http://tork-a.github.io/visualization_rwt/rwt_app_chooser/`

I can not access fetch1075's apps:
- `https://furushchev.ru/rwt_app_chooser/`
- `https://tork-a.github.io/visualization_rwt/rwt_app_chooser/`

I cannot access the whole web page:
- `https://localhost:8000/rwt_app_chooser/` (with locally `roslaunch rwt_app_chooser sample.launch`)
- `https://fetch15's IP address:8000/rwt_app_chooser/`
- `https://fetch1075's IP address:8000/rwt_app_chooser/`

708yamaguchi:
This is just a guess, but the reason we could not access tork-a.github.io is that it was redirected from http:// to https://.

So in this PR, we do not need to change the url.
It may only be necessary to just write a note to use http:// to README.md.

708yamaguchi:
@knorth55 
If you have time, could you check if you are in the same situation as me, where you can access the site via http:// but not via https://.

knorth55:
@708yamaguchi I confirmed and I got the same result!.
We need to use `http` to avoid this issue.

tkmtnt7000:
> After re-investigating, I think that http:// access has no problem, but https:// has a problem.
> I check several rwt_app_chooser and access ws://fetch1075.jsk.imi... with incognito chrome (with no cache mode)

I have confirmed this, too. But, I wonder why we access from https....

Affonso-Gui:
I have had similar problems with scratch. The project only supports `http`, but sometimes the browser insists in going to the `https` version.

In my experience:
- If the `http` part is omitted it will go to the `https`
- If you have accessed `https` once the browser will cache it and complete to `https` even  when explicitly typing `http`

708yamaguchi:
Thank you very much for your advice!

I update README to access rwt_app_chooser site via http, not https.
[`d58cecb` (#1529)](https://github.com/jsk-ros-pkg/jsk_robot/pull/1529/commits/d58cecb4f7a9ee80f9079423821079e2681f6891)

PR #1540, opened 2022-07-21
(baxtereus) Add roseus_resume initialization in baxter-interface.l

knorth55:
this PR adds `roseus_resume` initialization in `baxter-interface`.
with this commit, we can use `roseus_resume` in baxter, just by using `baxter-interface` or its child class.

cc. @Affonso-Gui


jsk_baxter_robot/baxtereus/baxter-interface.l
```diff
@@ -1,5 +1,6 @@
 (require "package://pr2eus_moveit/euslisp/robot-moveit.l")
 (require :baxter "package://baxtereus//baxter-util.l")
+(require :roseus_resume "package://roseus_resume/euslisp/interruption-handler.l")
```
pazeshun, line 3: I think this line causes error when `roseus_resume` is not installed in workspace.
Fix this or fix every .rosinstall to include `roseus_resume`.

jsk_baxter_robot/baxtereus/baxter-interface.l
```diff
@@ -47,6 +48,12 @@
      (ros::spin-once)
      (if mvit-rb (setq moveit-robot mvit-rb))
      (if mvit-env (send self :set-moveit-environment (send mvit-env :init :robot moveit-robot)))
+     (warning-message 3 "Installing interruption handler...~%")
+     (roseus_resume:install-interruption-handler self
```
pazeshun, line 52: Same as the above comment.

pazeshun (changes requested):

knorth55:
I tried to load `roseus_resume` when `ros::rospack-find "roseus_resume"` returns `t`.
```diff
 (require "package://pr2eus_moveit/euslisp/robot-moveit.l")
 (require :baxter "package://baxtereus//baxter-util.l")
-(require :roseus_resume "package://roseus_resume/euslisp/interruption-handler.l")
+(if (ros::rospack-find "roseus_resume")
+  (require :roseus_resume "package://roseus_resume/euslisp/interruption-handler.l"))
 (ros::load-ros-package "baxter_core_msgs")

 (defparameter *suction* 65537) ;;vacuum
```

But `roseus_resume` is the package name, so it causes error when `roseus_resume` is not found.

```
[rospack] Error: package 'roseus_resume' not found
Call Stack (max depth: 20):
/home/knorth55/install/jskeus/eus/Linux64/bin/irteusgl: ERROR th=0 no such package "ROSEUS_RESUME""ROSEUS_RESUME" in #<compiled-code #X55817cd305b0>E:
```

Are there any good trick to avoid this package name error?

pazeshun:
How about using macro?

knorth55:
> How about using macro?

No it does not work.
I think this error occurs during file read.
maybe here https://github.com/euslisp/EusLisp/blob/master/lisp/c/reader.c#L700 ?

```diff
 (require "package://pr2eus_moveit/euslisp/robot-moveit.l")
 (require :baxter "package://baxtereus//baxter-util.l")
-(require :roseus_resume "package://roseus_resume/euslisp/interruption-handler.l")
+(if (ros::rospack-find "roseus_resume")
+  (require :roseus_resume "package://roseus_resume/euslisp/interruption-handler.l"))
 (ros::load-ros-package "baxter_core_msgs")
 
 (defparameter *suction* 65537) ;;vacuum
@@ -49,10 +50,10 @@
      (if mvit-rb (setq moveit-robot mvit-rb))
      (if mvit-env (send self :set-moveit-environment (send mvit-env :init :robot moveit-robot)))
      (warning-message 3 "Installing interruption handler...~%")
-     (roseus_resume:install-interruption-handler self
-        right-gripper-action
-        left-gripper-action)
-     (roseus_resume:install-default-intervention self)
+     (eval '(roseus_resume:install-interruption-handler self
+               right-gripper-action
+               left-gripper-action))
+     (eval '(roseus_resume:install-default-intervention self))
      (warning-message 3 "...done.~%")
      ))
   (:right-property-cb (msg)

```

Affonso-Gui:
> Are there any good trick to avoid this package name error?

Probably the standard way is to use `#+` and `#-`.
In euslisp we need https://github.com/euslisp/EusLisp/pull/342 for this to bypass `package not found` errors (it is included in eus10, but not in the apt version).

A simple hack for it to work on apt eus as well is to define a dummy package:
```lisp
(if (ros::rospack-find "roseus_resume")
  (require :roseus_resume "package://roseus_resume/euslisp/interruption-handler.l"))
  (make-package "ROSEUS_RESUME"))
```
and then reference to internal symbols, i.e. double colons.
 `roseus_resume:install-interruption-handler` -> `roseus_resume::install-interruption-handler`

knorth55:
>and then reference to internal symbols, i.e. double colons.
roseus_resume:install-interruption-handler -> roseus_resume::install-interruption-handler

Thank you! I didn't know the difference between `:` and `::`.

Affonso-Gui (changes requested):
This will work with either `roseus_resume + eus10` or `no roseus_resume + eus9`, but will probably fail with `roseus_resume + eus9`. 
I have added eus10 `*features*` at https://github.com/Affonso-Gui/EusLisp/commit/ac5fd3581cb85dd68a4518430c67be0f2bb7a1d8, maybe relying on it could provide better coverage for all scenarios.

PR #1548, opened 2022-07-26
Pr/make init function for tweet client

sktometometo:
navigation-utils in jsk_fetch_startup load tweet_client.l from jsk_robot_startup. And it will wait until `/active_user/elapsed_time` parameter is available. This will sometimes prevent euslisp script execution when the parameter is not set.

This PR will fix it.

PR #1552, opened 2022-08-02
[jsk_robot_startup] Fix username retrieval in active_user.l

sktometometo:
Fix the problem reported in https://github.com/jsk-ros-pkg/jsk_robot/pull/1548.

Sometimes `active_user.l` fails to retrieve a username. This PR fix it.

### DIFF
--- a/jsk_robot_common/jsk_robot_startup/lifelog/active_user.l
+++ b/jsk_robot_common/jsk_robot_startup/lifelog/active_user.l
@@ -81,6 +81,7 @@
     (with-open-file
         (f (format nil "/tmp/username_~d.txt" (unix::getpid)))
       (setq username (read-line f)))
+    (unix::unlink "/tmp/username_~d.txt" (unix::getpid))
     username
     )
   )
@@ -95,6 +96,7 @@
     (with-open-file
         (f (format nil "/tmp/fullname_~d.txt" (unix::getpid)))
       (setq fullname (read-line f)))
+    (unix::unlink "/tmp/fullname_~d.txt" (unix::getpid))
     fullname
     )
   )
--- a/jsk_robot_common/jsk_robot_startup/lifelog/active_user.l
+++ b/jsk_robot_common/jsk_robot_startup/lifelog/active_user.l
@@ -75,7 +75,7 @@
   (let (username)
     (unix::system
      (format nil
-             "getent passwd `whoami` | cut -d ':' -f 1 > /tmp/username_~d.txt"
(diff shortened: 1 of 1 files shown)


knorth55:
in `/etc/password`, the curent `active_user.l` uses `user description`.
this PR changes to use `username`.

For example, in my case, `Shingo Kitagawa` is parsed with current master,
but this PR changes to use `knorth55`.
```
knorth55:x:1000:1000:Shingo Kitagawa:/home/knorth55:/bin/bash

```

k-okada:
So prefeered behavir is 
- if we have null "user description", then use "user name" ?

knorth55:
> So prefeered behavir is
if we have null "user description", then use "user name" ?

Yes, I think it is better.

jsk_robot_common/jsk_robot_startup/lifelog/active_user.l
```diff
@@ -71,6 +71,34 @@
...
+  (let (username)
+    (unix::system
+     (format nil
+             "whoami > /tmp/username_~d.txt"
+             (unix::getpid)))
+    (warn "Get PID ~A" (unix::getpid))
+    (with-open-file
+        (f (format nil "/tmp/username_~d.txt" (unix::getpid)))
+      (setq username (read-line f)))
+    username
```
k-okada, line 84: I think we should remove this generated file, `(unix::unlink "/tmp/username...")` ??  or use `(defun get-username () lisp::*user*)`

jsk_robot_common/jsk_robot_startup/lifelog/active_user.l
```diff
@@ -71,6 +71,34 @@
...
+    (unix::system
+     (format nil
+             "getent passwd `whoami` | cut -d ':' -f 5 | cut -d ',' -f 1 > /tmp/fullname_~d.txt"
+             (unix::getpid)))
+    (warn "Get PID ~A" (unix::getpid))
+    (with-open-file
+        (f (format nil "/tmp/fullname_~d.txt" (unix::getpid)))
+      (setq fullname (read-line f)))
+    fullname
+    )
```
k-okada, line 99: please remove `/tmp/fullname_...` too

k-okada (changes requested):

sktometometo:
Working on CI fixing

k-okada:
https://github.com/stack-of-tasks/eigenpy/pull/307 is blocker

PR #1556, opened 2022-08-07
[softhand_ros] use lambda-closure for subscriber

knorth55:
I tested with `eus10`, but `eus9` causes segmentation fault at `subscriber` in `let` scope.
I avoid the issue by using `lambda-closure`.

cc. @Affonso-Gui


Affonso-Gui:
Classic closure bug. Glad it worked on eus10!

PR #1557, opened 2022-08-08
[jsk_pr2_accessories] Add long fingertips used for @kosuke55's demos

kosuke55:
Added long fingertips that I used in my paper.
Sorry for being late to add them and forgetting the details of the files.

@Kanazawanaoaki @iory 
Could you please delete the ones you don't want or rename them appropriately?


iory:
@kosuke55 Thanks for your PR!

@Kanazawanaoaki Could you check the CAD files?

Kanazawanaoaki:
@kosuke55  
Thanks for your quick action! I have confirmed that all CAD files can be opened in solidworks and their contents seem to be OK.  

We decided to put all CAD related data on the google drive and only keep some used gripper data on github. Therefore, I would like you to change to add only `PR2_ChopStick_0_20200825_ver2` used in @kosuke55's thesis and `PR2_LongGripper_20191204` used in @ktro2828 's thesis to `jsk_robot/jsk_pr2_robot/jsk_pr2_accessories/`, which is the same hierarchy as `PR2_Gripper_CL-M3-5.0-BR_20191030`. 
I have already uploaded all the CAD and STL files you added [here](https://drive.google.com/drive/folders/1Ogv4izCKyPjHPOQjV_3YP9jTfbyXlvGj?usp=sharing ) on the shared drive.

@pazeshun If the original CAD file (i.e., `.SLDPRT` file) is too large, we could put only the STL file on github and write a link to the original CAD file in the README. What should we do this time?  
This time, `PR2_ChopStick_0_20200825_ver2.SLDPRT` is 163KB, `PR2_LongGripper_20191204.SLDPRT` is 153KB, and `PR2_Gripper_CL-M3-5.0-BR_20191030 SLDPRT`, which is already in the master, is 109KB.

pazeshun:
> This time, `PR2_ChopStick_0_20200825_ver2.SLDPRT` is 163KB, `PR2_LongGripper_20191204.SLDPRT` is 153KB

In my opinion, their file size seems OK to place here.

Perhaps we already have many original (and sometimes heavy) CAD files in `jsk_pr2_accessories`, so it's too late to worry about file size...

Kanazawanaoaki:
Thank you!

> Perhaps we already have many original (and sometimes heavy) CAD files in `jsk_pr2_accessories`, so it's too late to worry about file size...

I see ...

I will try to add the gripper used in the master's thesis demo as planned.

Kanazawanaoaki (approved):
I think this will be OK!

PR #1561, opened 2022-08-17
[jsk_fetch_startup] Enable shutdown_unchecked topic to prevent shutdown when battery is charging

iory:
# What is this?

This PR prevents shutdown if the robot is manually docked after kitchen demo has failed.

cc: @708yamaguchi @nakane11

### DIFF
--- a/jsk_fetch_robot/jsk_fetch_startup/apps/go_to_kitchen/go_to_kitchen.app
+++ b/jsk_fetch_robot/jsk_fetch_startup/apps/go_to_kitchen/go_to_kitchen.app
@@ -121,7 +121,7 @@ plugins:
     type: app_publisher/rostopic_publisher_plugin
     plugin_args:
       stop_topics:
-        - name: "/shutdown"
+        - name: "/shutdown_unchecked"
           pkg: std_msgs
           type: Empty
           cond:
--- a/jsk_fetch_robot/jsk_fetch_startup/launch/fetch_bringup.launch
+++ b/jsk_fetch_robot/jsk_fetch_startup/launch/fetch_bringup.launch
@@ -31,6 +31,9 @@
     </rosparam>
   </node>
 
+  <include file="$(find jsk_fetch_startup)/launch/fetch_shutdown.launch" >
+  </include>
+
   <node pkg="jsk_robot_startup" name="nav_speak" type="nav_speak.py" respawn="true" >
     <rosparam>
       lang: japanese
--- a/jsk_fetch_robot/jsk_fetch_startup/launch/fetch_shutdown.launch
+++ b/jsk_fetch_robot/jsk_fetch_startup/launch/fetch_shutdown.launch
@@ -1,0 +1,24 @@
+<launch>
+
+  <arg name="shutdown_topic" default="/shutdown" />
+  <arg name="shutdown_unchecked_topic" default="/shutdown_unchecked" />
+  <arg name="shutdown_dummy_topic" default="/shutdown_dummy" doc="Dummy input (no signal)." />
+  <arg name="battery_state_topic" default="/battery_state" />
+
+  <node name="input_shutdown_mux"
+        pkg="topic_tools" type="mux"
+        respawn="true" clear_params="true"
+	      args="$(arg shutdown_topic) $(arg shutdown_unchecked_topic) $(arg shutdown_dummy_topic)" >
+    <remap from="mux" to="input_shutdown_mux" />
+  </node>
+
+  <node name="input_shutdown_selector"
+        pkg="jsk_robot_startup" type="mux_selector.py"
+	      respawn="true"
+	      args="$(arg battery_state_topic) 'm.is_charging is True' $(arg shutdown_dummy_topic)
+              $(arg battery_state_topic) 'm.is_charging is False' $(arg shutdown_unchecked_topic)" >
+    <remap from="mux" to="input_shutdown_mux" />
+    <param name="wait" value="true" />
+  </node>
+
+</launch>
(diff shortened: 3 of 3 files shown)


708yamaguchi:
#### Problem

If fetch fails the kitchen demo, it automatically shuts down after a timeout.
This prevents the fetch from being left uncharged and degrading the battery.

The problem is that even if we manually return fetch to the dock before the kitchen demo times out, the shutdown will still take place.
This node checks the state of charge before shutdown, so this problem can be prevented.

#### Similar case

In PR2 or fetch cmd_vel, we are using `/base_controller/command_unchecked` topic that does not check the charge status.
So I think it is reasonable to use the shutdown_unchecked topic for the shutdown function as well.
https://github.com/jsk-ros-pkg/jsk_robot/blob/25b9f6a44f7639624cf8859e80215e5020028fdc/jsk_pr2_robot/jsk_pr2_startup/jsk_pr2_warning/cable_warning.l#L71
https://github.com/jsk-ros-pkg/jsk_robot/blob/25b9f6a44f7639624cf8859e80215e5020028fdc/jsk_fetch_robot/jsk_fetch_startup/scripts/warning.py#L93

nakane11:
In the kitchen demo, each movement (e.g. move-to-kitchen-door-front) is executed maybe up to [3-trial](https://github.com/knorth55/jsk_robot/blob/00ff7f510ebd65710abafc4fbfb1d9d7914b8b33/jsk_fetch_robot/jsk_fetch_startup/euslisp/go-to-kitchen.l#L7) * [10-retry](https://github.com/jsk-ros-pkg/jsk_pr2eus/blob/4700cf84db91d3b8d6cb17a540c29add5fbfd656/pr2eus/robot-interface.l#L1477) times. 
Therefore, app timeout ([1200](https://github.com/knorth55/jsk_robot/blob/00ff7f510ebd65710abafc4fbfb1d9d7914b8b33/jsk_fetch_robot/jsk_fetch_startup/apps/go_to_kitchen/go_to_kitchen.app#L6)) could happen before all move base actions end in failure and then the state transitions to auto-dock.

After the demo has timed out, there is some time before fetch shuts down. 
Since [stop_plugin of converter or uploader](https://github.com/knorth55/jsk_robot/blob/00ff7f510ebd65710abafc4fbfb1d9d7914b8b33/jsk_fetch_robot/jsk_fetch_startup/apps/go_to_kitchen/go_to_kitchen.app#L177-L190) could takes a long time, we can manually return fetch to the dock before shutdown_plugin is executed.

So I think this PR is useful.

nakane11:
When `/input_shutdown_mux` receives the selected topic for the first time, following warning appears on my pc. 
```
[ WARN] [1661238338.446011582]: Advertising on topic [/shutdown] with an empty message definition.  Some tools (e.g. rosbag) may not work correctly.
```
I confirmed that `/shutdown` was not published for the first input message by `rostopic echo /shutdown`. 
From the second time receiving `/shutdown_unchecked`, `/shutdown` comes to be published and the warning didn't appear.
It happened even if another node subscribes `/shutdown`.

I haven't try this on real fetch and cannot find a solution yet, but we can see the same warnings in jsk-fetch-startup.log.
I wonder this bug is causing the failure of shutdown recently.

708yamaguchi:
Thank you for your debugging.
I think it depends on the time between advertise and publish.

In my environment, this PR works properly if I add `sleep(1);` in the following line.
https://github.com/ros/ros_comm/blob/2a92e3d3ccc43ce3e9eac1b4e76bb7450c564aed/tools/topic_tools/src/mux.cpp#L182

nakane11:
I see. Thank you for letting me know.

708yamaguchi:
This is dirty solution,but publishing /shutdown_unchecked or /shutdown_dummy in advance is one of the solution.

At the beginning of fetch_shutdown.launch,
- If fetch is charged, publish /shutdown_unchecked
- If fetch is not charged, publish /shutdown_dummy

By doing this, mux can advertise /shutdown without fetch's shutdown.

k-okada:
@708yamaguchi @iory sorry for late, I have been thinking if there exists simpler solutions(ex. https://linuxjf.osdn.jp/JFdocs/linus-lecture/design.html) 

1) Check charning state within  stutdown plugin https://github.com/knorth55/jsk_robot/blob/00ff7f510ebd65710abafc4fbfb1d9d7914b8b33/jsk_fetch_robot/jsk_fetch_startup/apps/go_to_kitchen/go_to_kitchen.app#L191
support toipc based condition
```
          cond:
            - timeout
            - failure
            - rostopic /charning (--filter 'm.data == foo')
```
2) Always do not call shutdown if charning.
 2-A) Modify [shutdown.py](https://github.com/jsk-ros-pkg/jsk_robot/pull/1426/files#diff-cf0bde4b686b024bc710bc3ca3c01e227134a66e783ccea5defe6a0c90eccc29) ,
 2-B) send topic to [power_management.py](https://github.com/jsk-ros-pkg/jsk_robot/pull/1426/files#diff-cf0fbd7e530c81f32f775f350fb42dde871b58920a9f564d3d0cd8a423368b8f) from shutdown_plugin and publish '/shutdown' after checking chrgning state, we will subscribe (/shutdown and force_shutdown).
 2-C) Create node to set null to `~shutdown_command` when we push dock button and restore it then we push undock button.

iory:
@k-okada 
Thanks for your comments.
I thought it would be good in general to modify shutdown.py so that it can check the value of whatever topic.
I created the PR to enable it. 
https://github.com/jsk-ros-pkg/jsk_robot/pull/1571

k-okada:
@iory can we close this PR ? I have merged #1571

iory:
Yes. The problem was solved.

PR #1563, opened 2022-08-18
add allow-other-keys in baxter-init

knorth55:
I add `allow-other-keys` for `baxter-init`.
this is useful to pass all args with `&rest args` from your custom `baxter-init` function.

### DIFF
--- a/jsk_baxter_robot/baxtereus/baxter-interface.l
+++ b/jsk_baxter_robot/baxtereus/baxter-interface.l
@@ -383,7 +383,7 @@
    ))
 
 
-(defun baxter-init (&key (safe t) (type :default-controller) (moveit t))
+(defun baxter-init (&key (safe t) (type :default-controller) (moveit t) &allow-other-keys)
   (let (mvit-env mvit-rb)
     (when moveit
       (setq mvit-env (instance baxter-moveit-environment))
--- a/jsk_baxter_robot/baxtereus/baxter-softhand-interface.l
+++ b/jsk_baxter_robot/baxtereus/baxter-softhand-interface.l
@@ -178,7 +178,7 @@
 
 (defun baxter-init
   (&key (safe t) (type :default-controller) (moveit t)
-        (lgripper :softhand) (rgripper :softhand))
+        (lgripper :softhand) (rgripper :softhand) &allow-other-keys)
   (let (mvit-env mvit-rb)
     (if moveit
       (progn


pazeshun:
I'm not sure why, but CI fails. Maybe just waiting and re-running is better?

knorth55:
Weird...

```
2022-08-18T11:04:00.2128935Z E: Unable to locate package ros-melodic-fetch-bringup
```
```
2022-08-18T12:04:09.5690501Z FAILURE:
2022-08-18T12:04:09.5690617Z 
2022-08-18T12:04:09.5690742Z                                                                                 
2022-08-18T12:04:09.5691142Z [/github/home/ros/ws_jsk_robot/src/jsk_robot/jsk_fetch_robot/jsk_fetch_startup/launch/fetch_bringup.launch]:
2022-08-18T12:04:09.5691426Z 
2022-08-18T12:04:09.5691549Z                                                                                 
2022-08-18T12:04:09.5691880Z 	cannot find package [fetch_bringup] for node [controller_reset.py]
2022-08-18T12:04:09.5692099Z 
2022-08-18T12:04:09.5692224Z                                                                                 
2022-08-18T12:04:09.5692559Z unable to find node [fetch_bringup/controller_reset.py]: fetch_bringup
```

https://build.ros.org/job/Mbin_uB64__fetch_bringup__ubuntu_bionic_amd64__binary/

Affonso-Gui:
> this is useful to pass all args with &rest args from your custom baxter-init function

Could you give us a more concrete example?

knorth55:
Like this.

https://github.com/knorth55/eus_vive/blob/d65364023353679bc77b8d49b63efe988d397c5e/euslisp/lib/baxter-interface.l#L8

Affonso-Gui:
```lisp
(apply 'baxter-init :allow-other-keys t args)
```
http://www.lispworks.com/documentation/HyperSpec/Body/03_dada.htm

knorth55:
@Affonso-Gui 
> (apply 'baxter-init :allow-other-keys t args)

It works, thank you!
but this trick requires to modify the user's custom code.
with this PR, we don't need to change the user's custom code.
So, I think it is better to add `&allow-other-keys` for `baxter-init` function.

Affonso-Gui:
I am not sure if I have a well-established position in this.

The base function is totally independent from its caller, so it does seem to make more sense to fix the client code instead. 
All of the other robot-init functions also don't support `&allow-other-keys`, so that is another reason to keep the base unchanged.

However, `&allow-other-keys` is more standard than `:allow-other-keys`, and adding it to the base function shouldn't really have any drawbacks.

pazeshun (approved):
@Affonso-Gui 's comment
> All of the other robot-init functions also don't support &allow-other-keys

is correct. 
However, those robot-init functions don't even have `&key`, so `baxter-init` can be a standard of robot-init functions with `&key`.
Therefore, adding `&allow-other-keys` is OK, I think.
In addition, I am supported by the thought
> adding it to the base function shouldn't really have any drawbacks

k-okada:
Test failing
```
  -------------------------------------------------
  <testsuite tests="7" failures="3" disabled="0" errors="0" time="13" name="AllTests">
    <testcase name="all-test" status="run" time="0" classname="all-test">
    </testcase>
    <testcase name="test-baxter-moveit-init" status="run" time="11" classname="test-baxter-moveit-init">
     <failure message="baxter-interface is in simulation mode" type="AssertionError">
  Test:(not (send *ri* :simulation-modep))
  Trace:"^[0m-up-timer)
  imer-job* is ((lambda nil (send #&amp;lt;baxter-interface #X558b5f514530&amp;gt; :robot-interface-simulation-callback)) lisp::c"
  Message:"baxter-interface is in simulation mode"
     </failure>
    </testcase>
    <testcase name="test-baxter-moveit-angle-vector" status="run" time="0" classname="test-baxter-moveit-angle-vector">
    </testcase>
    <testcase name="test-baxter-moveit-angle-vector-fast" status="run" time="0" classname="test-baxter-moveit-angle-vector-fast">
    </testcase>
    <testcase name="test-baxter-moveit-angle-vector-sequence" status="run" time="1" classname="test-baxter-moveit-angle-vector-sequence">
    </testcase>
    <testcase name="test-baxter-moveit-angle-vector-sequence-fast" status="run" time="1" classname="test-baxter-moveit-angle-vector-sequence-fast">
    </testcase>
    <testcase name="test-baxter-moveit-angle-vector-start-time" status="run" time="0" classname="test-baxter-moveit-angle-vector-start-time">
     <failure message="start-time is ignored. Traj finishes at 0.022" type="AssertionError">
  Test:(&gt; tm-diff 3)
  Trace:"^[0mmstart testing [test-baxter-moveit-angle-vector-start-time]
  t]
  obot-interface-simulation-callback)) lisp::c).^[0m
  )
  imer-job"
  Message:"start-time is ignored. Traj finishes at 0.022"
     </failure>
    </testcase>
    <testcase name="test-baxter-moveit-angle-vector-sequence-start-time" status="run" time="0" classname="test-baxter-moveit-angle-vector-sequence-start-time">
     <failure message="start-time is ignored. Traj finishes at 0.046" type="AssertionError">
  Test:(&gt; tm-diff 3)
  Trace:"^[0mmstart testing [test-baxter-moveit-angle-vector-sequence-start-time]
  ).^[0m
  )
  imer-job).^[0m
  art testing [test-baxter-moveit"
  Message:"start-time is ignored. Traj finishes at 0.046"
     </failure>
    </testcase>
  </testsuite>
```

PR #1565, opened 2022-08-21
[baxtereus] move end-coords for new softhand mount

knorth55:
I change softhand mount, so I move the end-coords for softhand a little forward.
this mount is for softhand and softhand V2, so we can easily change the both two grippers.

IMG_2507

### DIFF
--- a/jsk_baxter_robot/baxtereus/baxter-softhand-interface.l
+++ b/jsk_baxter_robot/baxtereus/baxter-softhand-interface.l
@@ -196,6 +196,11 @@
         (if safe
           (setq *baxter* (instance baxter-robot-safe :init))
           (setq *baxter* (instance baxter-robot :init)))
+        ;; move end-coords for softhand
+        (if (equal lgripper :softhand)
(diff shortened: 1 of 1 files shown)


Affonso-Gui:
Just for curiosity, what are the key points and motivations for the new mount design?

jsk_baxter_robot/baxtereus/baxter-softhand-interface.l
```diff
@@ -196,7 +196,13 @@
         (if safe
           (setq *baxter* (instance baxter-robot-safe :init))
           (setq *baxter* (instance baxter-robot :init)))
-        (send *baxter* :angle-vector (send *ri* :state :potentio-vector))))
+        (send *baxter* :angle-vector (send *ri* :state :potentio-vector))
```
pazeshun, line 199: Sorry for very minor comment, but I think this line should be after translating end-coords.
What do you think?

knorth55: umm, thats right.
I will change it.

pazeshun (approved):

PR #1572, opened 2022-08-26
[jsk_baxter_startup] change twitter image in baxter.launch

knorth55:
when `spherical_stereo` is enabled, use the camera topic for twitter.

### DIFF
--- a/jsk_baxter_robot/jsk_baxter_startup/baxter.launch
+++ b/jsk_baxter_robot/jsk_baxter_startup/baxter.launch
@@ -61,6 +61,10 @@
        doc="Deprecated. Please use launch_twitter instead." />
   <arg name="launch_twitter" default="$(arg launch_tweet)"
        doc="launch twitter" />
+  <arg if ="$(arg launch_spherical_stereo)"
+       name="twitter_image_topics" default="/$(arg spherical_left_camera_ns)/image_raw"/>
+  <arg unless ="$(arg launch_spherical_stereo)"
+       name="twitter_image_topics" default="/cameras/head_camera/image"/>
   <arg name="launch_time_signal" default="true"/>
   <arg name="launch_app_manager" default="true"/>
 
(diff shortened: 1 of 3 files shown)


jsk_baxter_robot/jsk_baxter_startup/baxter.launch
```diff
@@ -14,6 +14,10 @@
   <arg name="launch_spherical_kodak" default="false" />
   <arg name="launch_spherical_stereo" default="false" />
   <arg name="start_openni" default="false"/>
+  <arg name="realsense_camera_ns" default="realsense_torso" />
+  <arg name="spherical_camera_ns" default="kodak_head" />
+  <arg name="spherical_left_camera_ns" default="elp_head_left" />
+  <arg name="spherical_right_camera_ns" default="elp_head_right" />
```
pazeshun, line 20: Why do you enable to change camera names?
In my point of view, those names are used in launch file names (e.g., baxter_***realsense_torso***.launch) and argument names (e.g., launch_***realsense_torso***), so should not be changed.

knorth55: this is just for reuse of the value.
i will change it from default to value

pazeshun: OK, thank you.
I want to ask one more: is `REALSENSE_CAMERA_NS` good name?
Perhaps in future, baxter may have a second realsense camera.
In that case, that name seems confusing.

knorth55: i think it is ok, and when we have another one we can change the name.
but what is your recommendation?

pazeshun: Oh, now I understand these names are the same names as the arguments of included roslaunch.
Changing them also seems confusing.
OK, my question is not good.

knorth55: so, it is better not to use capital letters?

pazeshun: I think using capital letters is permissible.
You don't have to change anything more, I think.
Sorry for confusing.

pazeshun (approved):

PR #1575, opened 2022-08-26
[jsk_baxter_startup] fix typo: default -> value

knorth55:
this is my mistake.
there is a typo in `baxter_softhand.launch` and `baxter_default.launch`.


pazeshun (approved):
It is our misfortune that wrongly-used `default` instead of `value` causes no error in roslaunch execution...

PR #1578, opened 2022-08-27
Adopt Google Chat ROS to smach_to_mail and adding launch file for it

mqcmd196:
Same with https://github.com/k-okada/jsk_robot/pull/72 . 
Using Google Chat in `smach_to_mail.py` and adding launch file.
Example output
chat_notification


jsk_robot_common/jsk_robot_startup/scripts/smach_to_mail.py
```diff
@@ -133,8 +139,12 @@ def _status_cb(self, msg):
                 for x in self.smach_state_list[caller_id]:
                     rospy.loginfo(" - At {}, Active state is {}{}".format(x['TIME'], x['STATE'],
                                   "({})".format(x['INFO']) if x['INFO'] else ''))
-                self._send_mail(self.smach_state_subject[caller_id], self.smach_state_list[caller_id])
-                self._send_twitter(self.smach_state_subject[caller_id], self.smach_state_list[caller_id])
+                if self.use_mail:
```
k-okada, line 142: If you do not want to publish `/email` `/twitter` and `/google_chat` all time, how about write code to check if there are topic subscrribers, instead of asking users to manually set `use_email` manually.
```
if rospy.get_num_connections() > 0:
   rospy.publish("email")
```

k-okada (changes requested):

jsk_robot_common/jsk_robot_startup/scripts/smach_to_mail.py
mqcmd196: I think the node for handling email, twitter, chatting should be launched all time, and if we want to suppress the notification, the demo code or this smach notification system should do it.

tkmtnt7000: > I think the node for handling email, twitter, chatting should be launched all time, and if we want to suppress the notification, the demo code or this smach notification system should do it.

I am for this idea. We set default interfaces when robot starts, but robots should switch to better interfaces dependent on demo or behavior. For example, we can use `info` data for switching notification interface or embed the message which interface should be used in `user_data`, though I still don't know what is the better way so much..

tkmtnt7000: So, I think this implementation is almost OK and the function of switching to better interfaces may be our near future work.

knorth55: how about using dynamic reconfigure to change `parameters` dynamically?
`rosparam` only loaded at the beginning, so other node cannot change it.

or `user_data` can contain the info, too.
also I think we dont need to suppress the notification from the beginning.
When someone says it is annoying, we can think about it probably 🤣

mqcmd196: I'm looking into ways to set the importance of smach information

mqcmd196: > how about using dynamic reconfigure to change parameters dynamically?
> rosparam only loaded at the beginning, so other node cannot change it.
> 
> or user_data can contain the info, too.
> also I think we dont need to suppress the notification from the beginning.
> When someone says it is annoying, we can think about it probably rofl

Sorry I overlooked this comment... and I've just created suppressing node by setting `notification level` dyn param. Please check https://github.com/jsk-ros-pkg/jsk_robot/pull/1648

PR #1582, opened 2022-09-01
[softhand_ros] use lower torque limit to avoid breakdown

knorth55:
I set lower torque limit for softhand to avoid breakdown.

cc. @YUKINA-3252 @soonhyo


pazeshun:
Could you test this for a while?
I'm not sure this is really a good value.

knorth55:
OK, I will test it.
this value is used about one year ago.

pazeshun:
Is the root problem fixed by #1622?
If #1622 is enough, I think this PR should not be merged because this PR makes grasping looser.

knorth55:
i thinks this pr is also needed because grasping torque is a bit high.
With this value, baxter can grasp object properly.

pazeshun (approved):

PR #1588, opened 2022-09-08
[jsk_robot_startup] Not using yaml file to set email config like receiver mail address

tkmtnt7000:
In this PR, both `smach_to_mail.py` and `email_topic.py` stop using yaml file to set email config.

`email_topic.py` has been used for some months, so it is deprecated rather than completely dropping yaml config.

### DIFF
--- a/jsk_robot_common/jsk_robot_startup/scripts/email_topic.py
+++ b/jsk_robot_common/jsk_robot_startup/scripts/email_topic.py
@@ -21,7 +21,8 @@ import base64
 class EmailTopic(object):
     """
     This node sends email based on received rostopic (jsk_robot_startup/Email).
-    Default values can be set by using `~email_info`
+    Default ~sender_address and ~receiver_address can be set from rosparam
+    Default values can be set by using `~email_info`(DEPRECATED)
 
     The yaml file is like the following:
     subject: hello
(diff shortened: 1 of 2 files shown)


knorth55:
i dont see the reason why you dont want to use yaml file...

tkmtnt7000:
https://github.com/k-okada/jsk_robot/pull/71#issuecomment-1223458987
I reflected this comment.
At that time, I thought putting config file under `/var/lib/robot/` makes debugging a little difficult...
cc: @708yamaguchi

708yamaguchi:
In my understanding, using yaml files is not a problem, but placing them in `/var/lib` is the problem.

tkmtnt7000:
Thank you very much.
So... maybe I should restore the ability to read yaml files and set email address as well as `email_topic.py`.
I dropped supporting yaml file in `smach_to_mail.py`.

708yamaguchi:
Ideally, I would like to use the same interface (e.g., how to specify sender_address) between `email_topic.py` and `smach_to_mail.py`.
If another `hoge_to_mail.py` is added in the future, this interface will be used.

How about defining a new class that can receive email information in both rosparam and yaml?

knorth55:
the current implementation is ok, I think.
because we can load `rosparam` from `yaml`  like `<rosparam command="load" file="hoge.yaml />`.

My point is why you need to write the email address in launch?.
We can share the info and save it in one yaml.

tkmtnt7000:
> My point is why you need to write the email address in launch?.
We can share the info and save it in one yaml.

I see. Thank you for information. 
So I should not write email address in launch files directly. We can share this info with other nodes in one yaml.

PR #1589, opened 2022-09-08
[jsk_robot_startup] Speak when shut down or reboot

nakane11:
Duplicate https://github.com/knorth55/jsk_robot/pull/319

Tell users that robot is shutting down like boot sound.

### DIFF
--- a/jsk_robot_common/jsk_robot_startup/scripts/shutdown.py
+++ b/jsk_robot_common/jsk_robot_startup/scripts/shutdown.py
@@ -57,19 +57,28 @@ class Shutdown(object):
             '~shutdown_command', '/sbin/shutdown -h now')
         self.reboot_command = rospy.get_param(
             '~reboot_command', '/sbin/shutdown -r now')
+        self.shutdown_sentence = rospy.get_param(
+            '~shutdown_sentence', 'シャットダウンします。')
+        self.reboot_sentence = rospy.get_param(
+            '~reboot_sentence', '再起動します。')
+        self.lang = rospy.get_param('~lang', 'jp')
         self.volume = rospy.get_param('~volume', 1.0)
+        self.timeout_server = rospy.get_param('~timeout_server', 1.0)
+        self.timeout_result = rospy.get_param('~timeout_result', 10.0)
 
-    def speak(self, text, lang='jp'):
-        if self.client.wait_for_server(timeout=rospy.Duration(1.0)):
+    def speak(self, text):
+        timeout_server = rospy.Duration(self.timeout_server)
+        timeout_result = rospy.Duration(self.timeout_result)
+        if self.client.wait_for_server(timeout=timeout_server):
             msg = SoundRequest()
             msg.sound = SoundRequest.SAY
             msg.command = SoundRequest.PLAY_ONCE
             msg.volume = max(0, min(1, self.volume))
             msg.arg = text
-            msg.arg2 = lang
+            msg.arg2 = self.lang
             goal = SoundRequestGoal(sound_request=msg)
             self.client.send_goal(goal)
-            self.client.wait_for_result(timeout=rospy.Duration(10.0))
+            self.client.wait_for_result(timeout=timeout_result)
             return self.client.get_result()
 
     def callback(self, msg):
@@ -101,7 +110,7 @@ class Shutdown(object):
                 return
 
         rospy.loginfo('Shut down robot.')
-        self.speak('シャットダウンします。')
+        self.speak(self.shutdown_sentence)
         ret = os.system(self.shutdown_command)
         if ret != 0:
             rospy.logerr("Failed to call '$ {}'. Check authentication.".format(
@@ -109,7 +118,7 @@ class Shutdown(object):
 
     def reboot(self, msg):
         rospy.loginfo('Reboot robot.')
-        self.speak('再起動します。')
+        self.speak(self.reboot_sentence)
         ret = os.system(self.reboot_command)
         if ret != 0:
             rospy.logerr("Failed to call '$ {}'. Check authentication.".format(
--- a/jsk_robot_common/jsk_robot_startup/scripts/shutdown.py
+++ b/jsk_robot_common/jsk_robot_startup/scripts/shutdown.py
@@ -5,7 +5,10 @@ import os
 from importlib import import_module
 
 import rospy
-from sound_play.libsoundplay import SoundClient
+import actionlib
+from sound_play.msg import SoundRequest
+from sound_play.msg import SoundRequestGoal
+from sound_play.msg import SoundRequestAction
 from std_msgs.msg import Empty
 
 
@@ -34,7 +37,8 @@ class Shutdown(object):
 
     def __init__(self):
         rospy.loginfo('Start shutdown node.')
-        self.client_jp = SoundClient(sound_action='/robotsound_jp', blocking=True)
+        self.client = actionlib.SimpleActionClient(
+            '/robotsound_jp', SoundRequestAction)
         self.condition = rospy.get_param(
             '~input_condition', None)
         if self.condition is not None:
@@ -55,9 +59,18 @@ class Shutdown(object):
             '~reboot_command', '/sbin/shutdown -r now')
         self.volume = rospy.get_param('~volume', 1.0)
 
-    def speak(self, client, speech_text, lang='jp'):
-        client.say(speech_text, voice=lang, volume=self.volume, replace=False)
-        return client.actionclient.get_result()
+    def speak(self, text, lang='jp'):
+        if self.client.wait_for_server(timeout=rospy.Duration(1.0)):
+            msg = SoundRequest()
+            msg.sound = SoundRequest.SAY
+            msg.command = SoundRequest.PLAY_ONCE
+            msg.volume = max(0, min(1, self.volume))
+            msg.arg = text
+            msg.arg2 = lang
+            goal = SoundRequestGoal(sound_request=msg)
+            self.client.send_goal(goal)
+            self.client.wait_for_result(timeout=rospy.Duration(10.0))
+            return self.client.get_result()
 
     def callback(self, msg):
         if isinstance(msg, rospy.msg.AnyMsg):
@@ -88,7 +101,7 @@ class Shutdown(object):
                 return
 
         rospy.loginfo('Shut down robot.')
-        self.speak(self.client_jp, 'シャットダウンします。')
+        self.speak('シャットダウンします。')
         ret = os.system(self.shutdown_command)
         if ret != 0:
             rospy.logerr("Failed to call '$ {}'. Check authentication.".format(
@@ -96,7 +109,7 @@ class Shutdown(object):
 
     def reboot(self, msg):
         rospy.loginfo('Reboot robot.')
-        self.speak(self.client_jp, '再起動します。')
+        self.speak('再起動します。')
         ret = os.system(self.reboot_command)
         if ret != 0:
             rospy.logerr("Failed to call '$ {}'. Check authentication.".format(
--- a/jsk_robot_common/jsk_robot_startup/scripts/shutdown.py
+++ b/jsk_robot_common/jsk_robot_startup/scripts/shutdown.py
@@ -1,9 +1,11 @@
 #!/usr/bin/env python
+# -*- coding: utf-8 -*-
 
 import os
 from importlib import import_module
 
 import rospy
+from sound_play.libsoundplay import SoundClient
 from std_msgs.msg import Empty
 
 
@@ -32,7 +34,7 @@ class Shutdown(object):
 
     def __init__(self):
         rospy.loginfo('Start shutdown node.')
-
+        self.client_jp = SoundClient(sound_action='/robotsound_jp', blocking=True)
         self.condition = rospy.get_param(
             '~input_condition', None)
         if self.condition is not None:
@@ -51,6 +53,11 @@ class Shutdown(object):
             '~shutdown_command', '/sbin/shutdown -h now')
         self.reboot_command = rospy.get_param(
             '~reboot_command', '/sbin/shutdown -r now')
+        self.volume = rospy.get_param('~volume', 1.0)
+
+    def speak(self, client, speech_text, lang='jp'):
+        client.say(speech_text, voice=lang, volume=self.volume, replace=False)
+        return client.actionclient.get_result()
 
     def callback(self, msg):
         if isinstance(msg, rospy.msg.AnyMsg):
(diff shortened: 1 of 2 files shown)


iory (approved):

k-okada:
I think `shutdown.py` is a fundamental program and we do not want to raise any errors during execution. For example,,,
```
        self.client_jp = SoundClient(sound_action='/robotsound_jp', blocking=True)
```
What will happen if sound_client failed to play sound and keep blocking ?

Another solution may use Non-ROS layer to tell everyone to shutdown.
As far as I understand, pr2 uses rc.d mechanism to play sound beep on startup.
and if `rc0.d/K00endbootstartshutdownbeep.sh` ? `rc6.d/S80endshutdownbeep.sh` ? `rc6.d/K00endbootstartshutdownbeep.sh`???  works correctly, it should play beep sound.

One reason would be... they try to play sound after OS stopped the sound driver? or something. 
In that case, we could chnage/move file name.

https://linuc.org/study/knowledge/515/
https://www.2daygeek.com/6-commands-to-shutdown-halt-poweroff-reboot-the-linux-system/





```
$ find rc* | grep beep
rc0.d/S80endshutdownbeep.sh
rc0.d/K00endbootstartshutdownbeep.sh
rc1.d/S99endbootstartshutdownbeep.sh
rc2.d/S99endbootstartshutdownbeep.sh
rc3.d/S99endbootstartshutdownbeep.sh
rc4.d/S99endbootstartshutdownbeep.sh
rc5.d/S99endbootstartshutdownbeep.sh
rc6.d/S80endshutdownbeep.sh
rc6.d/K00endbootstartshutdownbeep.sh
rcS.d/S19startbootbeep.sh

applications@pr1040:/etc/rc0.d $ cat K00endbootstartshutdownbeep.sh
#!/bin/sh

if test -e /etc/slave; then
# Slave boot-start
    if (echo $0|grep 'S..startbootbeep.sh' > /dev/null); then
        beep -f 523.25 -l 250 -n -f 587.33 -l 250 -n -f 659.26 -l 250 -n -f 783.99 -l 500
# Slave boot-finish
    elif (echo $0|grep 'S..endbootstartshutdownbeep.sh' > /dev/null); then
        beep -f 659.26 -l 250 -n -f 783.99 -l 250 -n -f 880.00 -l 250 -n -f 1046.50 -l 500
# Slave shutdown-start
    elif (echo $0|grep 'K..endbootstartshutdownbeep.sh' > /dev/null); then
        beep -f 1046.50 -l 250 -n -f 880.00 -l 250 -n -f 783.99 -l 250 -n -f 659.26 -l 500
# Slave shutdown-finish
    elif (echo $0|grep 'S..endshutdownbeep.sh' > /dev/null); then
        beep -f 783.99 -l 250 -n -f 659.26 -l 250 -n -f 587.33 -l 250 -n -f 523.25 -l 500
    fi

else

# Master boot-start
    if (echo $0|grep 'S..startbootbeep.sh' > /dev/null); then
        beep -f 261.63 -l 250 -n -f 293.44 -l 250 -n -f 329.63 -l 250 -n -f 392.00 -l 500
# Master boot-finish
    elif (echo $0|grep 'S..endbootstartshutdownbeep.sh' > /dev/null); then
        beep -f 329.63 -l 250 -n -f 392.00 -l 250 -n -f 440.00 -l 250 -n -f 523.25 -l 500
# Master shutdown-start
    elif (echo $0|grep 'K..endbootstartshutdownbeep.sh' > /dev/null); then
        beep -f 523.25 -l 250 -n -f 440.00 -l 250 -n -f 392.00 -l 250 -n -f 329.63 -l 500
# Master shutdown-finish
    elif (echo $0|grep 'S..endshutdownbeep.sh' > /dev/null); then
        beep -f 392.00 -l 250 -n -f 329.63 -l 250 -n -f 293.44 -l 250 -n -f 261.33 -l 500
    fi

fi

applications@pr1040:/etc/rc0.d $ cat S80endshutdownbeep.sh
#!/bin/sh

if test -e /etc/slave; then
# Slave boot-start
    if (echo $0|grep 'S..startbootbeep.sh' > /dev/null); then
        beep -f 523.25 -l 250 -n -f 587.33 -l 250 -n -f 659.26 -l 250 -n -f 783.99 -l 500
# Slave boot-finish
    elif (echo $0|grep 'S..endbootstartshutdownbeep.sh' > /dev/null); then
        beep -f 659.26 -l 250 -n -f 783.99 -l 250 -n -f 880.00 -l 250 -n -f 1046.50 -l 500
# Slave shutdown-start
    elif (echo $0|grep 'K..endbootstartshutdownbeep.sh' > /dev/null); then
        beep -f 1046.50 -l 250 -n -f 880.00 -l 250 -n -f 783.99 -l 250 -n -f 659.26 -l 500
# Slave shutdown-finish
    elif (echo $0|grep 'S..endshutdownbeep.sh' > /dev/null); then
        beep -f 783.99 -l 250 -n -f 659.26 -l 250 -n -f 587.33 -l 250 -n -f 523.25 -l 500
    fi

else

# Master boot-start
    if (echo $0|grep 'S..startbootbeep.sh' > /dev/null); then
        beep -f 261.63 -l 250 -n -f 293.44 -l 250 -n -f 329.63 -l 250 -n -f 392.00 -l 500
# Master boot-finish
    elif (echo $0|grep 'S..endbootstartshutdownbeep.sh' > /dev/null); then
        beep -f 329.63 -l 250 -n -f 392.00 -l 250 -n -f 440.00 -l 250 -n -f 523.25 -l 500
# Master shutdown-start
    elif (echo $0|grep 'K..endbootstartshutdownbeep.sh' > /dev/null); then
        beep -f 523.25 -l 250 -n -f 440.00 -l 250 -n -f 392.00 -l 250 -n -f 329.63 -l 500
# Master shutdown-finish
    elif (echo $0|grep 'S..endshutdownbeep.sh' > /dev/null); then
        beep -f 392.00 -l 250 -n -f 329.63 -l 250 -n -f 293.44 -l 250 -n -f 261.33 -l 500
    fi

fi
applications@pr1040:/etc/rc5.d $ cat S99endbootstartshutdownbeep.sh
#!/bin/sh

if test -e /etc/slave; then
# Slave boot-start
    if (echo $0|grep 'S..startbootbeep.sh' > /dev/null); then
        beep -f 523.25 -l 250 -n -f 587.33 -l 250 -n -f 659.26 -l 250 -n -f 783.99 -l 500
# Slave boot-finish
    elif (echo $0|grep 'S..endbootstartshutdownbeep.sh' > /dev/null); then
        beep -f 659.26 -l 250 -n -f 783.99 -l 250 -n -f 880.00 -l 250 -n -f 1046.50 -l 500
# Slave shutdown-start
    elif (echo $0|grep 'K..endbootstartshutdownbeep.sh' > /dev/null); then
        beep -f 1046.50 -l 250 -n -f 880.00 -l 250 -n -f 783.99 -l 250 -n -f 659.26 -l 500
# Slave shutdown-finish
    elif (echo $0|grep 'S..endshutdownbeep.sh' > /dev/null); then
        beep -f 783.99 -l 250 -n -f 659.26 -l 250 -n -f 587.33 -l 250 -n -f 523.25 -l 500
    fi

else

# Master boot-start
    if (echo $0|grep 'S..startbootbeep.sh' > /dev/null); then
        beep -f 261.63 -l 250 -n -f 293.44 -l 250 -n -f 329.63 -l 250 -n -f 392.00 -l 500
# Master boot-finish
    elif (echo $0|grep 'S..endbootstartshutdownbeep.sh' > /dev/null); then
        beep -f 329.63 -l 250 -n -f 392.00 -l 250 -n -f 440.00 -l 250 -n -f 523.25 -l 500
# Master shutdown-start
    elif (echo $0|grep 'K..endbootstartshutdownbeep.sh' > /dev/null); then
        beep -f 523.25 -l 250 -n -f 440.00 -l 250 -n -f 392.00 -l 250 -n -f 329.63 -l 500
# Master shutdown-finish
    elif (echo $0|grep 'S..endshutdownbeep.sh' > /dev/null); then
        beep -f 392.00 -l 250 -n -f 329.63 -l 250 -n -f 293.44 -l 250 -n -f 261.33 -l 500
    fi

fi

```

jsk_robot_common/jsk_robot_startup/scripts/shutdown.py
```diff
@@ -32,7 +34,7 @@ class Shutdown(object):
 
     def __init__(self):
         rospy.loginfo('Start shutdown node.')
-
+        self.client_jp = SoundClient(sound_action='/robotsound_jp', blocking=True)
```
iory, line 37: Without this action, an error will occur, so it is better to avoid it.

jsk_robot_common/jsk_robot_startup/scripts/shutdown.py
```diff
@@ -51,6 +53,11 @@ def __init__(self):
             '~shutdown_command', '/sbin/shutdown -h now')
         self.reboot_command = rospy.get_param(
             '~reboot_command', '/sbin/shutdown -r now')
+        self.volume = rospy.get_param('~volume', 1.0)
+
+    def speak(self, client, speech_text, lang='jp'):
+        client.say(speech_text, voice=lang, volume=self.volume, replace=False)
```
iory, line 59: What happens if an action is not executed correctly?
You can also add a timeout to avoid errors if the action is not executed.

iory (changes requested):

jsk_robot_common/jsk_robot_startup/scripts/shutdown.py
nakane11: I set `blocking=False`(default) due to the reason in the second reply.
Now `SoundClient` uses publisher instead of ActionClient.  
If we want to wait for speaking to finish, ActionClient is better?

jsk_robot_common/jsk_robot_startup/scripts/shutdown.py
nakane11: With timeout, we can use ActionClient if client does not exist. 
However, currently [`wait_for_server()`](https://github.com/ros-drivers/audio_common/blob/27c81ad628430472ddad4e5d063a766c16bf4bfb/sound_play/src/sound_play/libsoundplay.py#L347) is fixed to infinite.

I'm thinking of adding timeout arg to `say()`.
([battery_warning.py](https://github.com/jsk-ros-pkg/jsk_robot/blob/34a129462403878c9b4c9c86c1aa9b38d506c143/jsk_fetch_robot/jsk_fetch_startup/scripts/battery_warning.py#L34) also uses `wait_for_server()`, but it seems to be meaningless...)

nakane11:
Used SimpleActionClient to wait for speaking to finish before shutting down

jsk_robot_common/jsk_robot_startup/scripts/shutdown.py
```diff
@@ -51,6 +56,20 @@ def __init__(self):
             '~shutdown_command', '/sbin/shutdown -h now')
         self.reboot_command = rospy.get_param(
             '~reboot_command', '/sbin/shutdown -r now')
+        self.volume = rospy.get_param('~volume', 1.0)
+
+    def speak(self, text, lang='jp'):
+        if self.client.wait_for_server(timeout=rospy.Duration(1.0)):
```
iory, line 62: How about parameterizing the timeout value?

jsk_robot_common/jsk_robot_startup/scripts/shutdown.py
```diff
@@ -51,6 +56,20 @@ def __init__(self):
...
+            msg = SoundRequest()
+            msg.sound = SoundRequest.SAY
+            msg.command = SoundRequest.PLAY_ONCE
+            msg.volume = max(0, min(1, self.volume))
+            msg.arg = text
+            msg.arg2 = lang
+            goal = SoundRequestGoal(sound_request=msg)
+            self.client.send_goal(goal)
+            self.client.wait_for_result(timeout=rospy.Duration(10.0))
```
iory, line 71: How about parameterizing the timeout value?

iory (changes requested):

jsk_robot_common/jsk_robot_startup/scripts/shutdown.py
nakane11: Thanks for your advice. I added parameters.

jsk_robot_common/jsk_robot_startup/scripts/shutdown.py
```diff
@@ -81,13 +104,15 @@ def shutdown(self, msg):
                 return
 
         rospy.loginfo('Shut down robot.')
+        self.speak('シャットダウンします。')
```
iory, line 107: Is there any possibility that non-Japanese speakers will use this program? How about making this a parameter as well?

iory (changes requested):

jsk_robot_common/jsk_robot_startup/scripts/shutdown.py
nakane11: I added parameters for non-Japanese and updated the document.
I tried it on real PR2 robot.

jsk_robot_common/jsk_robot_startup/scripts/shutdown.py
```diff
@@ -32,7 +37,8 @@ class Shutdown(object):
 
     def __init__(self):
         rospy.loginfo('Start shutdown node.')
-
+        self.client = actionlib.SimpleActionClient(
+            '/robotsound_jp', SoundRequestAction)
```
iory, line 41: Could you add a document for `/robotsound_jp'?

iory (changes requested):

jsk_robot_common/jsk_robot_startup/scripts/shutdown.py
nakane11: Thanks! I updated the document.

iory (approved):
This change adds the ability to speak, which is good because timeout is set and users not using robotsound_jp can safely use the shutdown node.

k-okada:
I still in favor of adding spaking ability on highe level such as https://github.com/knorth55/jsk_robot/pull/314/files (or more higher) or lower level such as `rc.d` as startup beep sound.

nakane11:
OK. I'll consider the best way.

nakane11:
@k-okada 
fetch cannot use `beep` because it does not have buzzer(maybe it comes with a case) and I checked Boot Beep is set to enabled in BIOS. 

I tried to use rc.d and it worked on my laptop. https://github.com/jsk-ros-pkg/jsk_robot/compare/master...nakane11:jsk_robot:rc-shutdown
However, it seemed possible that aplay could not be executed before alsa became unavailable.

So I used systemd service instead. https://github.com/jsk-ros-pkg/jsk_robot/pull/1628
Even if services fail, it automatically end(/etc/systemd/system.conf) and shutdown is safely executed.

k-okada:
moved to #1628

PR #1595, opened 2022-09-13
[develop/fetch] add jsk-switch-wifi supervisor conf

knorth55:
this PR adds `jsk-switch-wifi-outside` and `jsk-switch-wifi-73B2` supervisor conf.
you can press the button on supervisor GUI, and you can change the network configuration.
In the conf, the follwing command runs.

```bash
nmcli c up sanshiro-73B2
# or 
nmcli c up sanshiro-outside
```

cc. @nakane11

### DIFF
--- a/jsk_fetch_robot/jsk_fetch_startup/supervisor_scripts/jsk-switch-wifi-73b2.conf
+++ b/jsk_fetch_robot/jsk_fetch_startup/supervisor_scripts/jsk-switch-wifi-73b2.conf
@@ -1,0 +1,8 @@
+[program:jsk-switch-wifi-73b2]
+command=nmcli c up sanshiro-73B2
+stopsignal=TERM
+autostart=false
+autorestart=false
+stdout_logfile=/var/log/ros/jsk-switch-wifi.log
+stderr_logfile=/var/log/ros/jsk-switch-wifi.log
+user=root
--- a/jsk_fetch_robot/jsk_fetch_startup/supervisor_scripts/jsk-switch-wifi-outside.conf
+++ b/jsk_fetch_robot/jsk_fetch_startup/supervisor_scripts/jsk-switch-wifi-outside.conf
@@ -1,0 +1,8 @@
+[program:jsk-switch-wifi-outside]
+command=nmcli c up sanshiro-outside
+stopsignal=TERM
+autostart=false
+autorestart=false
+stdout_logfile=/var/log/ros/jsk-switch-wifi.log
+stderr_logfile=/var/log/ros/jsk-switch-wifi.log
+user=root


sktometometo (approved):

k-okada:
If outside means corridor between B2 and A3, that would be mistake of 情報システム室
@mqcmd196 could ask them if it is intended?

mqcmd196:
I think `sanshiro-outside` means the access point in front of Murakami lab or toilet. @nakane11 says the wifi becomes so slow when Fetch arrives in front of the stairs because of the connection of an access point in 73B2

knorth55:
network connection at the corridar in front of 73B2 is weak.

Spot cannot connect to the controller via sanshiro-73B2 wifi in front of 73B2,
which @mqcmd196 and I faced this problem in RSJ.

Fetch can connect to sanshiro-73B2 wifi but pretty slow,
and hardly connect to sanshiro-73B2 at the cross of stairs and corridar,
as @nakane11 mentioned.

I'm wondering if we don't need to specify the BSSID of sanshiro-73B2,
and Fetch switches to use sanshiro-Murakami-lab at the corridar.

But the first thing is to change the wifi manually, I think.

k-okada:
Could someone try
https://askubuntu.com/questions/137460/wireless-roaming-aggressiveness
?
I think the default setting of signal too low, and it try to keep
original connection until it goes very weak.
--
◉ Kei Okada

2022年9月13日(火) 22:26 Shingo Kitagawa ***@***.***>:

sktometometo:
Driver seems not to support aggressiveness configuration.

k-okada:
Which chipset and driver are using/used in the robot?


1 does chip support aggressive parameter that help wifi connection status ?
try on widnows
https://blogs.umass.edu/Techbytes/2017/10/17/802-11ag-whats-the-difference-between-2-4ghz-and-5-0ghz/
2. How about Ubuntu:20.04 ??
https://askubuntu.com/questions/1270931/wi-fi-connection-does-not-roam-on-ubuntu-18-04
but
https://bootpanic.com/need-information-on-support-for-wifi-roaming-on-linux-platform/
say it is working on 18.04 ?? not firmware level?????
  How about setting it on wpa lavel????
https://wiki.archlinux.org/title/wpa_supplicant#Roaming
3. There darty workaround
https://askubuntu.com/questions/425583/ubuntu-connect-drops-worked-for-a-while-then-started-dropping-again/425617#comment1620622_425617
?

--
◉ Kei Okada



2022年9月14日(水) 10:20 Koki Shinjo ***@***.***>:

sktometometo:
I will check it again.

knorth55:
so, @sktometometo what's your idea about this PR? should we merge this PR, or try another way to switch wifi like #1598 ?

sktometometo:
@knorth55 I think this PR is useful for users who want to use Fetch outside of 73B2. So we should merge this PR. But `sanshiro-outside` configuration is not always better than `sanshiro-73B2`, especially the robot is in 73B2. This is because Wi-Fi roaming policy is poor in default configuration. We already have a solution for it in [this issue](https://github.com/jsk-ros-pkg/jsk_robot/issues/1598). So after merging this PR, it may be better to configure a script for Wi-Fi roaming policy which enables to use Fetch in a single NetworkManager profile in 73B2 and outside of 73B2.

PR #1596, opened 2022-09-13
[develop/fetch] add new arduino udev rules

knorth55:
someone left the local change in `fetch1075`, so I commit.
maybe @nakane11 ? or @708yamaguchi ?

### DIFF
--- a/jsk_fetch_robot/jsk_fetch_startup/udev_rules/80-arduino.rules
+++ b/jsk_fetch_robot/jsk_fetch_startup/udev_rules/80-arduino.rules
@@ -1,3 +1,4 @@
 # You can access arduino attached to fetch via /dev/arduino
 # https://github.com/start-jsk/jsk_apc/blob/16c004c511e864478aa6581a04c5a023e5fde391/jsk_arc2017_baxter/udev/90-rosserial.rules
 SUBSYSTEM=="tty", ATTRS{idVendor}=="0403", ATTRS{idProduct}=="6001", SYMLINK+="arduino", MODE="0666"
+SUBSYSTEM=="tty", ATTRS{idVendor}=="2341", ATTRS{idProduct}=="0058", SYMLINK+="arduino", MODE="0666"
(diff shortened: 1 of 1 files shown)


708yamaguchi (commented):
It would be good to assign an original name to the device file.
For example,
```
SUBSYSTEM=="tty", ATTRS{idVendor}=="2341", ATTRS{idProduct}=="0058", SYMLINK+="nakane_hand", MODE="0666"
```

This is my mistake and not good code, but I don't remember the purpose of the following code.
```
SUBSYSTEM=="tty", ATTRS{idVendor}=="0403", ATTRS{idProduct}=="6001", SYMLINK+="arduino", MODE="0666"
```

knorth55:
@708yamaguchi it means that this is not your change, right?
so this change is made by @nakane11  ?

nakane11:
Sorry for the late replay. 
Today I changed it to connect arduino of my robot hand to fetch1075.

708yamaguchi:
I told her to add udev rule to fetch1075.
I am sorry I did not tell her to commit the file.

nakane11:
My device is not important for other users and might be replaced with other products.
Is it possible to put rules directly in /etc/udev/rules.d or gitignore rules with specific name?

708yamaguchi:
> My device is not important for other users

Others may not use your device, but I don't think adding the udev rule will cause any inconvenience to others.

> might be replaced with other products.

If so, it would be good to write a comment in the udev rule.
```
# The following device is for arduino nano in nakane hand version 1
# Please update idVendor and idProduct if you use a different microcontroller
```

> Is it possible to put rules directly in /etc/udev/rules.d or gitignore rules with specific name?

It is difficult problem, but I prefer to git manage the udev files for the robot.
This is because other users can check the fetch settings and trace the development history with git.
How about sending a pull request only to `develop/fetch` branch now, and then to `master` branch after the stable version nakane hand is developed?

nakane11:
I agree. Thank you for your instruction.
I send a PR https://github.com/knorth55/jsk_robot/pull/345

knorth55:
@708yamaguchi can you review this PR?

708yamaguchi (approved):

PR #1600, opened 2022-09-15
[jsk_pr2_startup/cable_warning.l] No warning for zero velocity

nakane11:
If L1 is pressed, `/joy_vel` is selected and cable_warning node speaks "電源ケーブルを抜いてください。".
It is undesirable when we want to command torso up (L1+△)/torso down (L1+×) with PR2 connected to the cable. 
- [PR2 teleoperation](https://github.com/jsk-ros-pkg/jsk_robot/tree/master/jsk_pr2_robot#teleoperation)
- [JSK safe teleop system](https://github.com/jsk-ros-pkg/jsk_robot/tree/master/jsk_robot_common/jsk_robot_startup#launchsafe_teleoplaunch)
- [pr2_teleop_general_joystick.cpp](https://github.com/PR2/pr2_apps/blob/melodic-devel/pr2_teleop_general/src/pr2_teleop_general_joystick.cpp)

### DIFF
--- a/jsk_pr2_robot/jsk_pr2_startup/jsk_pr2_warning/cable_warning.l
+++ b/jsk_pr2_robot/jsk_pr2_startup/jsk_pr2_warning/cable_warning.l
@@ -19,6 +19,13 @@
 
 (defun move-base-command-cb (msg)
   (cond
+   ((and (= (send msg :linear :x) 0)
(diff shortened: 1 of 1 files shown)


iory (approved):
This is a good change because pr2 will not speak when we don't intend to move base.

PR #1612, opened 2022-09-22
[jsk_unitree_startup/cross] Use sshpass instead of /usr/bin/sshpass

k-okada:
split from https://github.com/k-okada/jsk_robot/pull/69

Is this for MacOS ??? @iory


iory:
Yes.  In Mac, sshpass will be installed on `/opt/homebrew/bin`
```
 $ which sshpass
/opt/homebrew/bin/sshpass
```

PR #1621, opened 2022-09-28
[softhand_ros, jsk_baxter_startup] add Baxter + Softhand urdf

knorth55:
I add Softhand urdf model and Baxter + Softhand URDF model

# Baxter + Softhand (Whole)

## Visual model
Screenshot from 2022-09-28 20-47-40

## Collision model
Screenshot from 2022-09-28 20-47-47

# Right Softhand

## Visual model
Screenshot from 2022-09-28 20-48-13

## Collision model
Screenshot from 2022-09-28 20-48-18

# Left Softhand

## Visual model
Screenshot from 2022-09-28 20-48-47

## Collision model
Screenshot from 2022-09-28 20-48-53

### DIFF
--- a/.github/workflows/config.yml
+++ b/.github/workflows/config.yml
@@ -77,6 +77,7 @@ jobs:
 
       - name: Start X server
         run: |
+          if [[ "${{ matrix.CONTAINER }}" =~ "jskrobotics/ros-ubuntu:14.04" ]]; then exit 0; fi
           echo 'debconf debconf/frontend select Noninteractive' | sudo debconf-set-selections || echo "failing on ros-ubuntu is OK" # set non interactive tzdata https://stackoverflow.com/questions/8671308/non-interactive-method-for-dpkg-reconfigure-tzdata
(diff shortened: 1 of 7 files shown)


pazeshun:
It's too late to say, but STL files are very heavy (6.44MB) ...
Heavy files make `git clone` slower, but this cannot be fixed because the commit including these files are already merged and we cannot forcibly delete that commit.

knorth55:
it was better to lessen the polygon of the meshes...

k-okada:
I see, I could remove STL file from commits, even from logs
https://qiita.com/xxl/items/7e91fc101d957629dedf
I'm afraid this will cause error when you run "git pull origin/master" in
existing workspace, but probably you can fix by clone again ore "git reset
--hard origin/master".
So if people agree with this inconvenience and upload STL file, I can work
on it.

--
◉ Kei Okada



2022年10月1日(土) 20:07 Shingo Kitagawa ***@***.***>:

PR #1634, opened 2022-10-02
[jsk_fetch_startup] Use email address instead of yaml for smach_to_mail

tkmtnt7000:
This change depends on https://github.com/jsk-ros-pkg/jsk_robot/pull/1588 .
We use rosparam instead of yaml file to set email information.

### DIFF
--- a/jsk_fetch_robot/jsk_fetch_startup/CMakeLists.txt
+++ b/jsk_fetch_robot/jsk_fetch_startup/CMakeLists.txt
@@ -90,10 +90,24 @@ macro(configure_dialogflow_hotword_yaml dname)
   endif()
 endmacro(configure_dialogflow_hotword_yaml)
 
+macro(configure_email_topic_yaml rname)
+  cmake_host_system_information(RESULT FETCH_NAME QUERY HOSTNAME)
+  if (${FETCH_NAME} MATCHES ^${rname})
+    set(RECEIVER_MAIL ${rname})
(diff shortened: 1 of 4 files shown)


knorth55 (changes requested):
this is not good because, it only supports fetch1075.
how about generating email_topic yaml in CMakeLists.txt?
like this: 

https://github.com/jsk-ros-pkg/jsk_robot/blob/68ae4be417c94d9d397c123f8d6daa08ade5dcdb/jsk_fetch_robot/jsk_fetch_startup/CMakeLists.txt#L78-L91

https://github.com/jsk-ros-pkg/jsk_robot/blob/68ae4be417c94d9d397c123f8d6daa08ade5dcdb/jsk_fetch_robot/jsk_fetch_startup/CMakeLists.txt#L95-L96

tkmtnt7000:
Thank you very much for reviewing. I mistakenly set email address without thinking fetch1075 and fetch15.
I think it may be good to generate a email config file from CMakeLists.txt.

tkmtnt7000:
Sorry for force-pushing...
Config yaml file is generated when we build workspace.

knorth55 (approved):

k-okada:
@tkmtnt7000 @knorth55 
Sorry, I did not look the changes of this PR carefully, I originaly intend to use rosparam directly within launch file, instead of using yaml file, 

see  
https://github.com/k-okada/jsk_robot/commit/37b71cea2c79abd31398073709e86d4fb36a7489, may be this is too complex....

PR #1648, opened 2022-10-06
[smach_to_mail] Add NOTIFICATION_LEVEL 

mqcmd196:
Related to https://github.com/jsk-ros-pkg/jsk_robot/pull/1578
I added notification level parameters in smach_to_state for deciding whether to notify each medium according to the importance of the information. It sounds like deciding logging level, like DEBUG, INFO, WARN, ERROR, FATAL... for each medium. It is strange for using WARN, ERROR, FATAL in this context, so I use DEBUG, INFO, NOTICE, CRITICAL, EMERGENCY instead (c.f. https://success.trendmicro.com/dcx/s/solution/TP000086250?language=en_US SEVERITY LEVEL)


k-okada:
this is interesting, but could provide any good example/senario of using this new feature? 

( Is it important/impressive than https://github.com/k-okada/jsk_robot/pull/76#issuecomment-1229097339, enable smach to respond to replies from chat client? )

PR #1650, opened 2022-10-07
[jsk_robot_startup] Use /email_topic/mail_title for email title in kitchen-demo

tkmtnt7000:
Duplicates #1649.

This PR enables us to set the same email title between smach_to_mail.py and app_manager_utile/mail_notifier_plugin.
We can receive the same demo emails as one thread.

app_managerで管理されているデモのうち，プラグインからメール通知が行くものに関して，smach_to_mail側が同じタイトルに合わせられるようにした変更です．
キッチンデモ関連のメールが行われた時間ごとに一つのスレッドにまとまるので，ユーザー通知+開発者向け通知のような形として見られるようになります．

### DIFF
--- a/jsk_robot_common/jsk_robot_startup/scripts/smach_to_mail.py
+++ b/jsk_robot_common/jsk_robot_startup/scripts/smach_to_mail.py
@@ -167,6 +167,8 @@ class SmachToMail():
 
         if subject:
             email_msg.subject = subject
+        elif rospy.has_param("/email_topic/mail_title"):
+            email_msg.subject = rospy.get_param("/email_topic/mail_title")
         else:
             email_msg.subject = 'Message from {}'.format(rospy.get_param('/robot/name', 'robot'))
 
--- a/jsk_fetch_robot/jsk_fetch_startup/apps/go_to_kitchen/go_to_kitchen.app
+++ b/jsk_fetch_robot/jsk_fetch_startup/apps/go_to_kitchen/go_to_kitchen.app
@@ -113,6 +113,7 @@ plugins:
     plugin_args:
       mail_title: Fetch kitchen patrol demo
       use_timestamp_title: true
+      use_app_start_time: true
     plugin_arg_yaml: /var/lib/robot/fetch_mail_notifier_plugin.yaml
   - name: move_base_cancel_plugin
     type: app_publisher/rostopic_publisher_plugin


jsk_robot_common/jsk_robot_startup/scripts/smach_to_mail.py
```diff
@@ -167,6 +167,8 @@ def _send_mail(self, subject, state_list):
 
         if subject:
             email_msg.subject = subject
+        elif rospy.has_param("/email_topic/mail_title"):
```
k-okada, line 170: Is `/email_topic/mail_tilte` used for  app_manager? if so, I think other rosparam uses private namespace such as
```
        if rospy.has_param("~sender_address"):
```
so I think this line also follows this convention. Please also update https://github.com/jsk-ros-pkg/jsk_robot/tree/master/doc/jsk_robot_common/jsk_robot_startup (I also forget to add ~sender_address and others...)

tkmtnt7000: Yes, `/email_topic/mail_title` is set by `app_manager`.
So, you mean we should unify namespace rules such as `/email_topic/sender_address` instead of `~sender_address`?

k-okada:
> So, you mean we should unify namespace rules such as /email_topic/sender_address instead of ~sender_address?

No, I think we should use `~mail_title` instead of
`/email_topic/mail_title`, because smach_to_mail is not designed only
for app_manager,
--
◉ Kei Okada

--
◉ Kei Okada



2022年10月13日(木) 16:16 Naoto Tsukamoto ***@***.***>:

knorth55:
IMO
If you want to pass data from node A to node B, you should use `topic`.
(because `parameter` does not change quickly, can be changed by anyone and remain until roscore dies, which causes bugs.)
If the data is constant, and node A and node B read the same data, you can use `parameter`.

tkmtnt7000:
Thank you for advice. 
Yes... it is more natural to send the email title as a topic, since it is not static.
I also started to think if apps which use `mail_notifier_plugin` run in parallel, using rosparam in its current implementation may have unintended behavior.
https://github.com/knorth55/app_manager_utils/blob/4ef13f4cd769c9df386eed9dd0b4cd715b64f32e/app_notifier/src/app_notifier/mail_notifier_plugin.py

k-okada:
なるほど，そういう用途なら
https://github.com/jsk-ros-pkg/jsk_robot/blob/a6cc2bbac2fe487d9fa95e9388c9a34644836ab7/jsk_robot_common/jsk_robot_startup/scripts/smach_to_mail.py#L97-L100
でsmachの中にメール題の書くのが良いのでは？

PR #1651, opened 2022-10-07
update time_signal schedule and speak minute info

knorth55:
this PR change `time_signal` app to speak `minute` info, too.
I also chage the schedule a little bit.


mqcmd196 (approved):
LGTM

PR #1652, opened 2022-10-08
[jsk_pr2_startup] disable gdrive_server in pr2_gazebo.launch

nakane11:
pr2.launch starts pr2_gdrive_server.launch by default from https://github.com/jsk-ros-pkg/jsk_robot/pull/1392, but I don't think it is needed for simulation.
If `client_secrets.json` is not placed, the nodes keep respawning.


knorth55 (approved):

knorth55:
Thank you for your catch!

PR #1655, opened 2022-10-12
[develop/fetch] [fetcheus] do not load roseus_resume when roseus_resume is not found

knorth55:
this PR allow not to use `eus10` or `roseus_resume`.
We can load `fetch-interface` in `eus9` and not `roseus_resume` installed workspace.
I also updated the readme for installation.


knorth55:
@MiyabiTane with this PR, we dont need to install `roseus_resume` or `eus10` to use `fetch`.

this PR is good for semi students.

Affonso-Gui (approved):

PR #1659, opened 2022-10-12
[jsk_fetch_startup] update time_signal schedule and speak more info

knorth55:
this PR changes `time_signal` schedule and change the speaking info.

### DIFF
--- a/jsk_fetch_robot/jsk_fetch_startup/scripts/time_signal.py
+++ b/jsk_fetch_robot/jsk_fetch_startup/scripts/time_signal.py
@@ -19,6 +19,7 @@ class TimeSignal(object):
         self.client_jp = actionlib.SimpleActionClient(
             '/robotsound_jp', SoundRequestAction)
         self.now_time = datetime.now()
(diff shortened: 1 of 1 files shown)


k-okada:
Does two robots speak same comment at same time?
I think the method to coordinates who to notify people around the robot would be very interesting topic for those who are interested in robot comunicate each other (@tkmtnt7000 )
For example, when time is approaching
- oveserve people and robot, and understand who is talking to whom
- if no one is talking, then start talking?
- if someint is talking, let them to understand that I am start talking?
etc...

tkmtnt7000:
>Does two robots speak same comment at same time?

Yes. Fetch15 and Fetch1075 speak the same comment at the same time.

> For example, when time is approaching
> 
> observe people and robot, and understand who is talking to whom
> if no one is talking, then start talking?
> if someint is talking, let them to understand that I am start talking?
> etc...

I'll try some methods. I think two robots do not always have to speak same comments.

tkmtnt7000 (approved):

PR #1667, opened 2022-10-13
[jsk_robot_startup] Fix sending notification every times if demo stuck and refactor _stop_timer_cb

tkmtnt7000:
This PR fixes a bug that caused emails to continue to be sent at certain intervals when the demo failed and the state transition stopped.

And I refactored `_stop_timer_cb` function (e.g. Log level, code indent).


knorth55 (approved):

tkmtnt7000:
Thank you for reviewing!

PR #1672, opened 2022-10-15
Add diagnostics_rx_error_summarizer.py for rx error confirmation

Kanazawanaoaki:
I would like to add a program based on [an existing program by @kosuke55 ](https://gist.github.com/kosuke55/4bbfca95a2f367b5b8ff8f9c6b3f8dfa#file-ethtool_rx_error_analyzer-py ) to `jsk_robot` in order to check which device is causing the error when the rx error occurs as reported by https://github.com/jsk-ros-pkg/jsk_robot/issues/1671 and https://github.com/jsk-ros-pkg/jsk_robot/issues/794

When executed without arguments, only devices with errors are displayed and the results are saved to a csv file. The --unsave argument eliminates saving to the csv file, and --show_all_ports outputs the results of all ports to stdout.
```
$ rosrun jsk_pr2_startup diagnostics_rx_error_summarizer.py
0
[Devices with errors]
1679040592 : /Motors/EtherCAT Device #00 (WG014) RX Error Port 0: 103
--------------------------------------------
1
[Devices with errors]
1679040593 : /Motors/EtherCAT Device #00 (WG014) RX Error Port 0: 103
--------------------------------------------
2
[Devices with errors]
1679040594 : /Motors/EtherCAT Device #00 (WG014) RX Error Port 0: 103
--------------------------------------------
3
[Devices with errors]
1679040595 : /Motors/EtherCAT Device #00 (WG014) RX Error Port 0: 103
--------------------------------------------
```


708yamaguchi (commented):
Thanks for the nice script.
I am able to use this script with pr1012 today and there seems no problems.

It would be helpful if you could give a timestamp or number of attempts in the stdout of this script.

708yamaguchi:
@Kanazawanaoaki 
I think that timestamps and graph drawing functions are important to know when packet drops occur.
So it would be great if we could output csv or draw graphs by default.

jsk_pr2_robot/jsk_pr2_startup/jsk_pr2_warning/diagnostics_rx_error_summarizer.py
```diff
@@ -0,0 +1,87 @@
...
+        if self.is_csv:
+            print(self.count)
+            for s in status:
+                if 'values' not in dir(s):
+                    continue
+                for v in s.values:
+                    if 'RX Error' in v.key:
+                        if self.count == 0:
+                            self.keys.append(s.name + ' - ' + v.key)
+                            self.values.append(v.value)
```
nakane11, line 45: I think only timestamps are written into csv now.
```suggestion
                        self.values.append(v.value)
```
I would appreciate it if the timestamp is displayed on stdout, too.

Kanazawanaoaki:
Thanks for your comments. I have updated the default to save to a csv file, to save more than just the timestamp in the csv, and to display the timestamp in the stdout as well.

PR #1680, opened 2022-10-19
[WIP] [jsk_fetch_startup] Refactor navigation-utils.l

tkmtnt7000:
`navigation-utils.l` contains kitchen-demo specific function, so I devide it from `navigation-util.l` and move it to `go-to-kitchen.l`.
I think users easily uses the navigation function when `navigation-utils.l` has slim implementation.

**What I did**
- Remove dependency on switchbot_ros in `navigation-utils.l`
- Code related to state-machine moves to `go-to-kitchen.l`

Since the function is simply moved, there are ways to improve it a bit more, such as by making it a class.


knorth55:
please rebase this or check #1703

PR #1679, opened 2022-10-19
[jsk_fetch_startup] Fix description report-light-on

tkmtnt7000:
Thanks to https://github.com/knorth55/jsk_robot/pull/343#issuecomment-1283283835.

Fix description in `report-light-on`.

PR #1684, opened 2022-10-20
[WIP] [develop/fetch] [jsk_fetch_startup] Macro approach of #1679

knorth55:
I add `def-state-func` macro to reset `image` and `description in `userdata`.
This is a macro approach recommended by @k-okada 

WE DID NOT TESTED YET.

cc. @k-okada @tkmtnt7000


Affonso-Gui (changes requested):
This macro doesn't account for potential `&optional` and `&rest`
```lisp
(def-state-func foo (&optional userdata)
    (print 10))

(defun foo (&optional userdata)
   (set-alist 'description "" &optional)
   (set-alist 'image "" &optional)
   (print 10))
```

It is also worth reiterating that we will have bugs as soon as the user forgets to register functions with this macro. Some other ideas might be:
- Incorporating the `description` and `image` tags as an innate part of state machines, and resetting them in the `:exec-state-machine` method
-  Calling `:publish-status` manually from within the functions, either in a different namespace or filtering out the doubles in `smach_to_mail.py`

knorth55:
Hmm... it is pretty rare case to use `&rest` and `&optional` for this function
because the state-machine function always have one argument `userdata`.

knorth55:
I modified to support `&rest` and `&optional`.
[jsk-ros-pkg/jsk_robot@`29154d7` (#1684)](https://github.com/jsk-ros-pkg/jsk_robot/pull/1684/commits/29154d7f4dab6a88257f5230be3b0f264e029109)

Affonso-Gui (changes requested):
```lisp
(def-state-func foo (&optional (userdata nil))
    userdata)

(defun foo (&optional (userdata nil))
   (set-alist 'description "" (userdata nil))
   (set-alist 'image "" (userdata nil))
   userdata)
```

knorth55:
I support `(&optional (userdata nil))` case.
I also add test.

PR #1687, opened 2022-10-20
[develop/fetch] diable google chat when google_chat_space is not set

knorth55:
this PR diable `google_chat_ros` when `google_chat_space` param is not set.
currently, `fetch1075` does not send e-mail because of this.


tkmtnt7000 (approved):
Thank you for debugging.

mqcmd196 (approved):
LGTM

PR #1688, opened 2022-10-20
[jsk_robot_startup] disable google chat when google_chat_space is not set

knorth55:
`~google_chat_space` is necessary field for `smach_to_mail.py` if you want to use `Google Chat`
this PR check if `~google_chat_space` is set, and if not disable `Google Chat`.


tkmtnt7000 (approved):

mqcmd196 (approved):
LGTM

PR #1692, opened 2022-10-21
[WIP] [develop/pr2] fix check_openni_node.py

knorth55:
- fix volueme in `check_openni_node.py`
- comment out subprocess of `roslaunch`.

we are testing on pr1040


knorth55:
@Kanazawanaoaki

PR #1695, opened 2022-10-22
[jsk_robot_startup] Add timeout for smach notification

knorth55:
this PR adds timeout for smach notification.
when app failed with error and smach does not send any `finish` or `end` conditions, the node will send notifaction with this timeout.

### DIFF
--- a/jsk_robot_common/jsk_robot_startup/scripts/smach_to_mail.py
+++ b/jsk_robot_common/jsk_robot_startup/scripts/smach_to_mail.py
@@ -73,6 +73,10 @@ class SmachToMail():
             self._gchat_thread = None
 
     def _stop_timer_cb(self, event):
+        '''
+        If smach does not go to finish/end state,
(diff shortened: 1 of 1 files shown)


k-okada (commented):
that's good point, for example https://github.com/k-okada/jsk_robot/tree/master/jsk_naoqi_robot/jsk_pepper_startup/apps/meeting takes at least 3 hours. So in some cases, it is difficult to estimate duration.

One idea is to  check if the smach is working by checking node status? or some topics.

Related to meeting app, I also thinking to add option to send immediately or send when finished option.

tkmtnt7000:
Thank you for sending PR to master branch and thank you for advice.

>One idea is to check if the smach is working by checking node status? or some topics.

I think it may be good to check smach node status.

knorth55:
@tkmtnt7000 please open a PR to implement the feature.

PR #1699, opened 2022-10-22
[jsk_robot_startup] add def-state-func macro in smach-utils.l

knorth55:
for `smach_to_mail.py`, we need to initialize `description` and `image` for every state at the beginning.
so I add `def-state-func` macro to do it automatically.
I also add test to check this macro.

this is part of #1684 .

cc. @tkmtnt7000


jsk_robot_common/jsk_robot_startup/test/test-smach-utils.l
```diff
@@ -0,0 +1,29 @@
+(require :unittest "lib/llib/unittest.l")
+(require :smach-utils "package://jsk_robot_startup/euslisp/smach-utils.l")
+
+
+(defmacro def-state-func-test (func-name result)
+  `(deftest ,(intern (format nil "~A-test" func-name)) ()
+     (let* ((func '(,func-name '((description . "hoge") (image . "fuga"))))
+            (output (eval func)))
+       (assert (equal output ,result)
+               (format nil "~%func: ~A error, output: ~A~%" ,(format nil "~A" func-name) output)))))
```
Affonso-Gui, line 10: Another example with more checks.

```lisp
(defmacro def-state-func-test (func-name)
  `(deftest ,(intern (string-upcase (format nil "~A-test" func-name))) ()
     ,@(mapcar #'(lambda (out expected)
                   `(let ((output (,func-name ',out)))
                      (assert (equal output ',expected)
                              (format nil ,(format nil "~~%func: ~A error, output: ~~S~~%" func-name) output))))
               (list '((description . "hoge") (image . "fuga"))
                     '((description . "hoge") (image . "fuga") (other . "value"))
                     '((description . "hoge"))
                     nil)
               (list '((description . "") (image . ""))
                     '((description . "") (image . "") (other . "value"))
                     '((description . ""))
                     nil))))
```

- Prefer upcase strings with intern
- No need to use eval, we can just paste the code in
- Prefer ~S in assert formatting
```
~A: func: state-func error, output: ((description . ) (image . ))
~S: func: state-func error, output: ((description . "") (image . ""))
```

jsk_robot_common/jsk_robot_startup/euslisp/smach-utils.l
```diff
@@ -0,0 +1,15 @@
...
+                   ((listp (cadr args))
+                    (if (listp (caadr args))
+                      (cadr (caadr args))
+                      (caadr args)))
+                   (t (cadr args))))
+                ((eq (car args) '&rest) `(car ,(cadr args)))
+                (t (car args)))))
+    `(defun ,name ,args
+       (setf (cdr (assoc 'description ,userdata)) "")
+       (setf (cdr (assoc 'image ,userdata)) "")
```
Affonso-Gui, line 14: Avoid errors if the userdata doesn't have the fields yet.
```lisp
(if (assoc 'description ,userdata) (setf (cdr (assoc 'description ,userdata)) ""))
(if (assoc 'image ,userdata) (setf (cdr (assoc 'image ,userdata)) ""))
```

Affonso-Gui (changes requested):

jsk_robot_common/jsk_robot_startup/euslisp/smach-utils.l
```diff
@@ -0,0 +1,15 @@
...
+  (let ((userdata
+          (cond ((eq (car args) '&optional)
+                 (cond
+                   ((listp (cadr args))
+                    (if (listp (caadr args))
+                      (cadr (caadr args))
+                      (caadr args)))
+                   (t (cadr args))))
+                ((eq (car args) '&rest) `(car ,(cadr args)))
+                (t (car args)))))
```
Affonso-Gui, line 11: Maybe also `((null args) (error "Function must have the userdata parameter"))` ?

knorth55:
I updated.

PR #1705, opened 2022-10-25
[jsk_pr2_startup] fix check_openni_node.py

knorth55:
this PR fix the behavior of `check_openni_node` in `jsk_pr2_startup`.

- stop launching `kinect_head.launch` in `check_openni_node.py`
  - this causes zombies processes
  - this process is out of control from `pr2.launch`, so it causes name crash.
- kill `kinect_head_nodelet_manager` and other child processes
- set `respawn` `true` for `kinect_head_nodelet_manager`
  - so `check_openni_node` kill the nodelet manager, and nodelet manager and child processes automatically respawn.
 
this fix now works on `pr1040`, and now it correctly reset kinect head.
cc. @Kanazawanaoaki 

FYI:
- I copied and modified `manager.launch.xml` from `rgbd_launch`
  - similar PR is already open in https://github.com/ros-drivers/rgbd_launch/pull/40
- I copied and modifed `openni.launch` from `openni_camera`
  - I made a PR to add `manager_respawn` in https://github.com/ros-drivers/openni_camera/pull/78


jsk_pr2_robot/jsk_pr2_startup/jsk_pr2_sensors/openni.launch
```diff
@@ -0,0 +1,118 @@
+<!-- Entry point for using OpenNI devices -->
```
k-okada, line 1: if this is copied from  `(find openni_launch)/launch/openni.launch` please add commen in the top of this file and add what did you changed with reason.

k-okada (changes requested):

jsk_pr2_robot/jsk_pr2_startup/jsk_pr2_sensors/openni.launch
knorth55: I updated and added the comments

PR #1707, opened 2022-10-27
define delete-attached-object-by-id for pr2eus < 0.3.14

k-okada:


jsk_fetch_robot/fetcheus/fetch-interface.l
```diff
@@ -6,6 +6,14 @@
 (ros::load-ros-package "fetch_driver_msgs")
 (ros::load-ros-package "robot_controllers_msgs")
 
+(when (null (assoc :delete-object-by-id (send collision-object-publisher :methods)))
```
tkmtnt7000, line 9: I think the background of this feature may be important.
Something like the following.
```suggestion
(when (null (assoc :delete-object-by-id (send collision-object-publisher :methods)))
  ;; jsk-ros-pkg/jsk_pr2eus <0.3.14 does not have :delete-object-by-id method
  ;; See https://github.com/jsk-ros-pkg/jsk_demos/pull/1375#issuecomment-1292943531
```

tkmtnt7000 (approved):
Thank you for updating!

PR #1719, opened 2022-11-10
[jsk_robot_startup] Support manual setting of the name of mongodb database and collection

tkmtnt7000:
This PR supports launching multiple `common_logger.launch` .
Related to #1706 #1714

Use case:
- Divide database dependent on each demonstration. (kitchen demo, PR2 fridge demo, etc)
- Able to change database name and collection name.

cc: @knorth55

Issue #1714, opened 2022-11-01
Pepper stores compressed image on mongo


knorth55:
@k-okada 
pepper stores compressed image on mongodb  in musca, which will cause slow query soon.
This was the reason why I gave up using mongo at the past.
If you want to enable compressed image for all robot, we need a better computer with better storage.
Also, we need to run Mongo backup more frequently, which i have done manually.
what is your plan?

knorth55:
i think for mongo, it is better to store data periodically, like smach execution because mongo is slow and heavy.
now in smach status local_data,
we have image data, so we dont need to store the image all the time if you want to use smach execution image.

honestly, the mongo server will be slow and heavy soon, and i dont want to maintain the musca mongo server so often.

cc. @tkmtnt7000

k-okada:
> Also, we need to run Mongo backup more frequently, which i have done manually.

Oh, I didn't know that. Please explain how to backup mongo data, I am happy to maintain that. Beacause this is an essense of our furutre research direction and also I'd like to show how to utilize stored data in next demonstaration (assiged: @tkmtnt7000, @mqcmd196  )
You can also remove current pepper/basil data, if there are already some problem. It is just a testing.

> i think for mongo, it is better to store data periodically

periodically? on an as-needed basis ?? 
https://github.com/jsk-ros-pkg/jsk_recognition/pull/2736 stores data only when data has changed.

> now in smach status local_data, we have image data, so we dont need to store the image all the time if you want to use smach execution image.

I'd like to build memory of robot when @tkmtnt7000 and @a-ichikura talking each other and show some object based on their conversation in front of the robot. May be we can store good image sequence if we build smach which the state changes in very fast loop, but keep images other than smach state also make sense, I think.

> honestly, the mongo server will be slow and heavy soon, and i dont want to maintain the musca mongo server so often.

I agree, thus why I am waiting for https://github.com/jsk-ros-pkg/jsk_3rdparty/issues/372 and https://github.com/jsk-ros-pkg/jsk_robot/pull/1574.

But I understand that moving to new system is not as easy as we thought in the beginning, and I always beleive "worse is always better than worst".

I think one reason of heavy db is we store all data in to one collection, if this is true, one idea is to separate collection based on episode.

knorth55:
> Oh, I didn't know that. Please explain how to backup mongo data, I am happy to maintain that

[...]

Issue #1706, opened 2022-10-25
mongodbの使い方


k-okada:
@knorth55 以下のlaunchファイルを立ち上げたら，とりあえずDBに何か貯まるのかな？と思ったんだけど，
```
[ERROR] [1666693339.502586]: Could not get message store services. Maybe the message store has not been started? Retrying..
[ERROR] [1666693339.502586]: Could not get message store services. Maybe the message store has not been started? Retrying..
```
というエラーがでる．
１）ローカルPCだけに保存する場合
２）リプリケーとする場合
のそれぞれのデフォルトの設定の方法を教えてください．
```
<launch>
  <!-- logging database -->
  <include file="$(find jsk_robot_startup)/lifelog/mongodb.launch">
    <arg name="use_daemon" value="true"/>
    <arg name="port" value="27017" />
    <arg name="repl_set_mode" value="false" />
  </include>

  <arg name="map_frame" default="odom" />
  <include file="$(find jsk_robot_startup)/lifelog/common_logger.launch">
    <arg name="save_rgb" value="false" />
    <arg name="save_depth" value="false" />
    <arg name="save_tf" value="true" />
    <arg name="save_joint_states" value="true" />
    <arg name="save_base_trajectory" value="false" />
    <arg name="save_object_detection" value="false" />
    <arg name="save_action" value="true" />
    <arg name="save_smach" value="true" />
    <arg name="save_app" value="true" />
    <arg name="enable_monitor" value="false" />
    <arg name="log_rate" value="1.0" />
    <arg name="launch_manager" value="true" />
    <arg name="map_frame_id" value="$(arg map_frame)" />
    <arg name="approximate_sync" value="true"/>
  </include>
</include>
```

k-okada:
thank you!
It sees mongodb is working
```
k-okada@p51s:~/catkin_ws/ws_hironx/src/rtmros_tutorials/hrpsys_ros_bridge_tutorials$ cd
k-okada@p51s:~$ sudo systemctl status mongodb.service
[sudo] password for k-okada:
● mongodb.service - An object/document-oriented database
   Loaded: loaded (/lib/systemd/system/mongodb.service; enabled; vendor preset: enabled)
   Active: active (running) since Tue 2022-10-25 19:16:32 JST; 2h 55min ago
     Docs: man:mongod(1)
 Main PID: 6030 (mongod)
    Tasks: 28 (limit: 4915)
   CGroup: /system.slice/mongodb.service
[...]

PR #1720, opened 2022-11-10
[jsk_robot_startup] Support manual setting of the name of mogodb database and collection

tkmtnt7000:
This is a master version of #1719

---
This PR supports launching multiple `common_logger.launch` .
Related to https://github.com/jsk-ros-pkg/jsk_robot/issues/1706 https://github.com/jsk-ros-pkg/jsk_robot/issues/1714

Use case:

Divide database dependent on each demonstration. (kitchen demo, PR2 fridge demo, etc)
Able to change database name and collection name.

### DIFF
--- a/jsk_robot_common/jsk_robot_startup/lifelog/common_logger.launch
+++ b/jsk_robot_common/jsk_robot_startup/lifelog/common_logger.launch
@@ -1,4 +1,8 @@
 <launch>
+  <!-- mongo setting -->
+  <arg name="database" default="jsk_robot_lifelog" />
+  <arg name="collection" default="null" />
+
   <!-- options -->
   <arg name="save_rgb" default="true" />
   <arg name="save_depth" default="true" />
@@ -33,8 +37,9 @@
   <machine if="$(arg localhost)" name="localhost" address="localhost" />
 
   <!-- nodelet -->
+  <arg name="ns" default="lifelog" />
   <arg name="launch_manager" default="true" />
-  <arg     if="$(arg launch_manager)" name="manager" default="mongodb_record_nodelet_manager" />
+  <arg     if="$(arg launch_manager)" name="manager" default="mongodb_$(arg ns)_nodelet_manager" />
   <arg unless="$(arg launch_manager)" name="manager" />
 
   <!-- others -->
@@ -53,7 +58,7 @@
         pkg="nodelet" type="nodelet" args="manager" machine="$(arg machine)"
         output="screen" respawn="$(arg respawn)" launch-prefix="$(arg launch-prefix)"/>
 
-  <group ns="lifelog">
+  <group ns="$(arg ns)">
     <!-- image throttle -->
     <group if="$(arg save_rgb)">
       <group if="$(arg save_depth)">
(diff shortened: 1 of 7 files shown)


jsk_robot_common/jsk_robot_startup/lifelog/common_logger.launch
```diff
@@ -184,8 +205,12 @@
           respawn="$(arg respawn)"
           if="$(arg save_tf)">
       <rosparam subst_value="true">
+        database: $(arg database)
         update_rate: $(arg log_rate)
       </rosparam>
+      <rosparam subst_value="true" if="$(eval arg('collection') != '')">
```
k-okada, line 211: @tkmtnt7000 can we avoid using `eval` ??
which is available from kinetic http://wiki.ros.org/roslaunch/XML

tkmtnt7000: I understand, (but it seems that `eval` makes codes easier to read...)

One solution is that we set rosparam `collection` manually  in each demo launch file instead of `arg`, like the following.

```xml
<include file="$(find jsk_robot_startup)/lifelog/common_logger.launch">
  <arg name="hoge" value="hoge" />
  ...
  <param name="collection" value="go_to_kitchen" />
</include>
```

tkmtnt7000: Thank you very much for suggestion.
I updated in [559f602](https://github.com/jsk-ros-pkg/jsk_robot/pull/1720/commits/559f602efcc67fcefbba4ff07f1982501aefdbda),[fda2d1a](https://github.com/jsk-ros-pkg/jsk_robot/pull/1720/commits/fda2d1af52e18afa2de4d573c7a86961443dfac2), [648157c](https://github.com/jsk-ros-pkg/jsk_robot/pull/1720/commits/648157ca858568b493dd4389f285228ae3673cea)

k-okada (commented):
do we need to update https://github.com/jsk-ros-pkg/jsk_robot/blob/master/jsk_pr2_robot/jsk_pr2_startup/jsk_pr2_warning/battery_logger.py and jsk_pr2_startup/pr2.launch too?

do we need to change `/robot/database` to `~robotdatabase` ? if we use `/robot/database`, we do not have to update lauch fiels?

tkmtnt7000:
Sorry for late response.
>do we need to update [master/jsk_pr2_robot/jsk_pr2_startup/jsk_pr2_warning/battery_logger.py](https://github.com/jsk-ros-pkg/jsk_robot/blob/master/jsk_pr2_robot/jsk_pr2_startup/jsk_pr2_warning/battery_logger.py?rgh-link-date=2023-01-24T08%3A32%3A38Z) and jsk_pr2_startup/pr2.launch too?

I think it is No because `battery_logger.py` does not seem to use mongodb, and this PR's change does not have an effect on exsiting launch files.

>do we need to change /robot/database to ~robotdatabase ? if we use /robot/database, we do not have to update lauch fiels?

I think it is no, too. Fetch and PR2 use /robot/database now.

PR #1736, opened 2022-11-18
[develop/fetch] camera shutter volume control ad sound only when speak is t 

knorth55:
dont play shutter sound when `speak` is `nil`.
we also control the shutter sound by `volume`.


sktometometo (approved):

sktometometo:
Sorry for late reviewing

PR #1741, opened 2022-11-22
[fetcheus] go-pos and go-pos-unsafe test for kinematics simulation

knorth55:
this PR checks if we can run `:go-pos` and `:go-pos-unsafe` in `fetch` + `kinematics simulator`.
this test will fail, but #1740 fix the bug.


k-okada:
closed via #1740

PR #1749, opened 2022-12-02
[panda_eus] check franka_description exists in panda_eus

knorth55:
this PR fixes build error with `franka_description` old version.
`franka_description` 's version is lower than 0.10.0, the directory structure is different and `franka_description/robots/common` and `franka_description/robots/panda`  do not exist.
https://github.com/frankaemika/franka_ros/tree/0.9.1/franka_description/robots

this PR check if the directory exists or not.

### DIFF
--- a/jsk_panda_robot/panda_eus/CMakeLists.txt
+++ b/jsk_panda_robot/panda_eus/CMakeLists.txt
@@ -13,7 +13,17 @@ catkin_package()
 ###
 ### dual_panda.l generation
 ###
-if(franka_description_FOUND AND (NOT ("$ENV{ROS_DISTRO}" STRLESS "melodic")))
+
+if(EXISTS ${franka_description_SOURCE_PREFIX}/robots)
+  set(_franka_urdf ${franka_description_SOURCE_PREFIX}/robots)
+else()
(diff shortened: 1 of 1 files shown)


pazeshun:
Sorry for the inconvenience, but is this correct way to fix?
I think this type of problem always occur, and our usual strategy is just writing version information to `package.xml`, isn't it?
https://github.com/jsk-ros-pkg/jsk_robot/blob/2614319f42403d4d09441cf9e395ff872494f48c/jsk_fetch_robot/jsk_fetch_startup/package.xml#L14
Currently, version information is not written, this is my fault.
So I think what I have to do is to write version information to package.xml.
What do you think?

pazeshun:
Or perhaps do you have any reason that you cannot upgrade `franka_description`?

knorth55:
i think we can add both.
checking version is also good, but we need to run `rosdep` again.
i dont know `rosdep` force to upgrade the version.

pazeshun:
OK, could you check #1751 ?
I think just checking `franka_description` version is simpler, easier to read, and easier to deal with future updates of `franka_description`.

knorth55:
lets move to #1751.

PR #1751, opened 2022-12-04
[panda_eus] Skip generating dual_panda.l when franka_description or xacro is old to avoid build failure

pazeshun:
Another solution of #1749 
~~Currently, building `panda_eus` fails when `franka_description`'s version is less than 0.10.0.~~
Currently, building `panda_eus` fails when `franka_description`'s version is less than 0.10.0 or `xacro`'s version is less than 1.13.14.
This PR fixes it by skipping `dual_panda.l` generation.

### DIFF
--- a/jsk_panda_robot/panda_eus/CMakeLists.txt
+++ b/jsk_panda_robot/panda_eus/CMakeLists.txt
@@ -13,15 +13,20 @@ catkin_package()
 ###
 ### dual_panda.l generation
 ###
-if(franka_description_FOUND AND (NOT ("$ENV{ROS_DISTRO}" STRLESS "melodic")))
-  # xacro.load_yaml cannot be recognized under melodic, while it is recommended on melodic and upper.
-  # PR introducing xacro.load_yaml: https://github.com/ros/xacro/pull/283
-  # Related issue: https://github.com/ros/xacro/issues/298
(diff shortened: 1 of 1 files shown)


knorth55 (approved):

knorth55:
can you add the version information in `package.xml` like this?
this changes nothing now, but it may change the behavior in the future.

:eyes: 
https://github.com/ros-infrastructure/rosdep/pull/694

```
 <build_depend version_gte="0.7.11">fetch_teleop</build_depend> 
```

pazeshun:
From my point of view, writing version to both CMakeLists.txt and package.xml seems not easy to maintenance, because these versions should be synchronized.
We can leave comments, but they may be ignored.
Also, the future version_gte may cause build failure
when the version is not met.
What you want to accomplish with #1749 is not to cause build failure, isn't it?

knorth55:
what i want to do is avoiding the build failure.
but imo, it is not hard to maintain both, but it depend on your decision.

knorth55:
@pazeshun can you also check https://github.com/jsk-ros-pkg/jsk_robot/issues/1758 , too?

pazeshun:
> @pazeshun can you also check #1758 , too?

Maybe that issue was fixed by the latest commit.

knorth55 (approved):

PR #1752, opened 2022-12-05
[develop/fetch]  move app_schedule.yaml in jsk_fetch_startup

knorth55:
this PR moves `app_schedule.yaml` in `jsk_fetch_startup` and manage the file by git.


tkmtnt7000 (approved):
That's good change. I think it becomes easier to catch changes in these yaml files.

PR #1753, opened 2022-12-05
[develop/fetch] use enable and speak_enable for volume control

knorth55:
this PR set `enable` and `speak_enable` to `nil` for `volume zero` app.
setting `volume` `0` means that robot speaks but no voice.
this causes interruption when other nodes speaks something.

this PR changes to set `enable` and `speak_enable` `nil`.
this change does not sent `sound_play` action, so it does't interrupt any other nodes.


knorth55:
@Kanazawanaoaki can you review this PR?

Kanazawanaoaki (approved):
I didn't consider interrupting when other nodes speak.  
LGTM!

PR #1774, opened 2023-02-10
PR2 Noetic Branch

knorth55:
PR2 Noetic branch

Now PR1040 is using this branch.


tkmtnt7000 (commented):
Just a quick look, but I may not think pr1040 is publishing `speech_to_text`.

knorth55:
`/speech_to_text` is published from `/speech_to_text_mux`.
`mux` does not register publisher until the topic arrived.

tkmtnt7000:
Thank you. I understand.

tkmtnt7000:
That's why we have to speak to PR2, "ねえねえ" twice when we are going to launch applications from dialogflow task executive.

knorth55:
yes.
and the problem can be solved by this PR.

https://github.com/ros/ros_comm/pull/2305

PR #1785, opened 2023-03-24
[jsk_robot_startup] advertise when no tweet publisher is advertised yet

knorth55:
advertise publisher needed for tweet in `tweet-string` when the publishers are not yet advertised.
with this change, we don't need to write `(ros::advertise "/tweet" ...)` in our script, and we can tweet by just runining `tweet-string`

### DIFF
--- a/jsk_robot_common/jsk_robot_startup/lifelog/tweet_client.l
+++ b/jsk_robot_common/jsk_robot_startup/lifelog/tweet_client.l
@@ -6,7 +6,10 @@
 
 (load "package://pr2eus/speak.l")
 
-(defun tweet-string (twit-str &key (warning-time) (with-image) (image-wait 30) (speak t))
+(defun tweet-string (twit-str &key (warning-time) (with-image) (image-wait 30) (speak t)
+                              (tweet-topic-name "/tweet"))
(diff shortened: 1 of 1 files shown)


tkmtnt7000:
Nice change!

jsk_robot_common/jsk_robot_startup/lifelog/tweet_client.l
```diff
@@ -7,6 +7,10 @@
 (load "package://pr2eus/speak.l")
 
 (defun tweet-string (twit-str &key (warning-time) (with-image) (image-wait 30) (speak t))
+  (if (null (ros::get-num-subscribers "/tweet"))
+    (ros::advertise "/tweet" std_msgs::String 1))
+  (if (null (ros::get-num-subscribers "/tweet_image_server/tweet/goal"))
+    (ros::advertise "/tweet_image_server/tweet/goal" rostwitter::TweetActionGoal 1))
```
tkmtnt7000, line 13: Almost same as above.

knorth55: the origin/master branch does not use `rostwitter` message, so I removed this change

tkmtnt7000 (approved):

PR #1788, opened 2023-03-26
[jsk_robot_startup] fix python3 decode error in smach_to_mail

knorth55:
this PR fix error in `python3` + `smach_to_mail.py`.
`x['MAIL']` is `bytes` object, so it should be decoded to `utf-8`.


tkmtnt7000 (commented):
Thank you very much for debugging.

Although it may not be very relevant to this error, it may be a good idea to write the info and description in the same way like this.
https://github.com/tkmtnt7000/jsk_robot/commit/e09738a6c7fdab041619049c55dc3553433cd311

knorth55:
can you open a pr to update this pr?

tkmtnt7000:
Sure. Um.. send pr to jsk-ros-pkg directly?? i have just sent pr to knorth55's branch.

PR #1793, opened 2023-04-26
[WIP] [unitree/cross] Fix unitree system build

tkmtnt7000:
Currently, unitree system build (`build_ros1_dependencies.sh` and `build_ros1.sh`) fails at several points.

This PR includes
- [x]  Fix `app_manager`  url and version (772dc9a67ad1d15795a24fee85848da2c0f5cdc1)
- [x] Install python-typing to avoid import typing error (772dc9a67ad1d15795a24fee85848da2c0f5cdc1)
- [x] Add libfcl-dev and libccd-dev to dependencies because `geometric_shapes` package needs them (0ec465c1b8483f221ccfb971108eaefca41a6ac2, [929ae6e](https://github.com/jsk-ros-pkg/jsk_robot/pull/1793/commits/929ae6e28c9ad51e28228155123d939f9540999d))
- [x] Avoid setuptools version error caused by `spidev==3.6` (af35715082126af9c475411e4f3bb292c6595127)

But, this PR  do not fix the following issue

- [ ] Unable to install `python-twisted` in `build_ros1_dependencies.sh` because of the following error. 
```
Installed /opt/jsk/System/ros1_dependencies/lib/python2.7/site-packages/Twisted-17.9.0-py2.7-linux-aarch64.egg                                                                                                                                                                                                                                                                               
Processing dependencies for Twisted==17.9.0                                                                                                                                                                                                                                                                                                                                                  
Searching for zope.interface>=3.6.0                                                                                                                                                                                                                                                                                                                                                          
Reading https://pypi.python.org/simple/zope.interface/                                                                                                                                                                                                                                                                                                                                       
Downloading https://files.pythonhosted.org/packages/1f/ef/cae21b6295cb2960429f463e6e99daf3623a0c4ced7242a809a0c94c9605/zope.interface-6.1a2.tar.gz#sha256=167ea1a5c28aad2fa7fb0ca8761229bf54766ab6e696c80e592ee8c936ab1776                                                                                                                                                                   
Best match: zope.interface 6.1a2                                                                                                                                                                                                                                                                                                                                                             
Processing zope.interface-6.1a2.tar.gz                                                                                                                                                                                                                                                                                                                                                       
Writing /tmp/easy_install-ur5jc9/zope.interface-6.1a2/setup.cfg                                                                                                                                                                                                                                                                                                                              
[...]

PR #1794, opened 2023-04-27
 [unitree/cross] Fix unitree system build

k-okada:
#1793 tried to use lateset python modules, but it seems too difficult, 
This PR try to work with same version of python modules that has been working before.


tkmtnt7000 (commented):
I have confirmed that we can execute `make system` successfully by this PR. (but i have not confirmed that source file generated by this pr really works on real robot yet.)

k-okada:
@tkmtnt7000 please verify whether this PR correctly works on the actual robot

tkmtnt7000:
Sorry for late reply.

I have confirmed that this PR's branch works successfully on a real robot (go1 pro ver) with lead_teleop as a minimum demonstration.


(I think that https://github.com/iory/jsk_robot/tree/unitree-camera-walk-app, https://github.com/k-okada/jsk_robot/pull/66, and so on may be necessary to execute richer demo such as using google chat)

PR #1314, opened 2021-01-06
[jsk_pr2_startup] add virtual_force_publisher to pr2.launch

knorth55:
launch `virtual_force_publisher` in `pr2.launch` by default.

cc. @ishiguroJSK

PR #1795, opened 2023-04-28
[jsk_pr2_startup] Change launch_virtual_force arg to avoid dupulicated node error

nakane11:
Currently both `pr2.launch` and `pr2_bringup.launch` run virtual_force_publisher and the nodes are killed for `new node registered with same name` error.
This PR makes  `pr2.launch` stop launching the nodes.

### DIFF
--- a/jsk_pr2_robot/jsk_pr2_startup/pr2_bringup.launch
+++ b/jsk_pr2_robot/jsk_pr2_startup/pr2_bringup.launch
@@ -159,9 +159,6 @@
     </include>
   </group>
 
-  <!-- Virtual force -->
-  <include file="$(find virtual_force_publisher)/launch/dualarm_virtual_force_publisher.launch" />
-
   <!-- Audio -->
   <include file="$(find audio_play)/launch/play.launch" if="$(arg launch_audio_play)">
     <arg name="ns" value="audio_play" />


Kanazawanaoaki (approved):
I hadn't noticed. Nice PR!
LGTM.

k-okada:
What was the reason for needing the commit at https://github.com/jsk-ros-pkg/jsk_robot/commit/ffc10275694e046a1b01e1e3030da56dc47d6043 (or https://github.com/jsk-ros-pkg/jsk_robot/pull/1251/files) in the first place? Is there any issue if we revert it?" @Kanazawanaoaki

Kanazawanaoaki:
I'm sorry. I just remembered.  
The reason for this problem is that the same node was added to different files in the two PRs https://github.com/jsk-ros-pkg/jsk_robot/pull/1251 (in `pr2_bringup.launch` launch by `robot` ) and https://github.com/jsk-ros-pkg/jsk_robot/pull/1314 (in `pr2.launch`  launch by `jsk-pr2-startup` ), and we created https://github.com/jsk-ros-pkg/jsk_robot/pull/1314 without considering the possibility that https://github.com/jsk-ros-pkg/jsk_robot/pull/1251 would be merged and left it as it was.  

(It seems that only https://github.com/jsk-ros-pkg/jsk_robot/pull/1314 had been included in the `develop/pr2` branch, but both were included at the time of the noetic transition, etc., causing the problem.)

Kanazawanaoaki:
I don't remember why it was added to `pr2.launch`, in the [PR]( https://github.com/jsk-ros-pkg/jsk_robot/pull/1314) of the newly created one, but I think it was said that virtual_force_publisher sometimes crashes, ,so it is better to use jsk-pr2-startup, which is easier to start up.

@nakane11 
Therefore, if there is no particular reason, it may be better to delete virtual_force_publisher in pr2_bringup.launch and leave pr2.launch.

nakane11:
@Kanazawanaoaki 
Thank you for reviewing. I deleted virtual_force_publisher in pr2_bringup.launch and rebased the commits.

Kanazawanaoaki:
Thank you very much. I think it will be OK this time.

PR #1208, opened 2020-02-27
[jsk_fetch_startup] add fetch etc files

sktometometo:
- add upstart scripts and launch files which are usually in /etc/init/ and /etc/ros/indigo.
- add a script to deploy the upstart scripts
- modify the upstart scripts so that they will use jsk_fetch_startup version of launch files instead of robot.launch in /etc/ros/indigo

PR #1799, opened 2023-05-09
add single panda

yuki-asano:
単腕のpandaを追加したものです。

Screenshot from 2023-05-09 16-07-56


jsk_panda_robot/panda_eus/CMakeLists.txt
```diff
@@ -34,14 +34,30 @@ if(franka_description_FOUND
...
   message(WARNING "Dependency is not met, so skip generating dual_panda.l")
   message(WARNING "franka_description version: ${franka_description_VERSION}, must be >= ${_franka_description_min_ver}")
   message(WARNING "xacro version: ${xacro_VERSION}, must be >= ${_xacro_min_ver}")
 endif()
 
 
+###
+### panda.l generation
+###
+add_custom_command(OUTPUT ${PROJECT_SOURCE_DIR}/models/panda.l
```
pazeshun, line 48: Please move `add_custom_command` and `add_custom_target` for `panda.l` into the same `if` block as `dual_panda.l` generation, because you cannot generate `panda.l` outside that `if` block.
Please read https://github.com/jsk-ros-pkg/jsk_robot/blob/93919243019ae06be43874022062abbeee00349e/jsk_panda_robot/panda_eus/CMakeLists.txt#L21-L27
Maybe you have to change some messages:
`dual_panda.urdf.xacro assumes` -> `panda.urdf.xacro and dual_panda.urdf.xacro assume` in
https://github.com/jsk-ros-pkg/jsk_robot/blob/93919243019ae06be43874022062abbeee00349e/jsk_panda_robot/panda_eus/CMakeLists.txt#L22
`dual_panda.l` -> `panda.l and dual_panda.l` in
https://github.com/jsk-ros-pkg/jsk_robot/blob/93919243019ae06be43874022062abbeee00349e/jsk_panda_robot/panda_eus/CMakeLists.txt#L39

jsk_panda_robot/panda_eus/euslisp/panda-interface.l
```diff
@@ -0,0 +1,237 @@
...
+          )
+  )
+(defmethod panda-robot-interface
+  (:init
+   (&rest args &key ((:controller-timeout ct) nil))
+   (prog1
+       (send-super* :init :robot panda-robot
+                    :joint-states-topic "joint_states"
+                    :controller-timeout ct args)
+     ;; check whather the environement is Gazebo
```
pazeshun, line 21: typo: `whather` -> `whether`

jsk_panda_robot/panda_eus/euslisp/panda-interface.l
```diff
@@ -0,0 +1,237 @@
...
+     (setq gazebop (numberp (ros::get-param "/gazebo/time_step")))
+     (if gazebop (ros::ros-warn "Using Gazebo environment"))
```
pazeshun, line 23: Have you ever tested gazebo environment?

jsk_panda_robot/panda_eus/euslisp/panda-interface.l
```diff
@@ -0,0 +1,237 @@
...
+         arm
+       (:arm (setq acts (list gripper-move-action)))
+       )
+     (when acts
+       (dolist (act acts)
+         (send self :send-gripper-move-action act
+               pos (/ (* 1000 0.08) tm) :wait nil))
+       (if wait (mapcar #'(lambda (act) (send act :wait-for-result)) acts))
+       )))
+  )
```
pazeshun, line 225: Many parts are common to https://github.com/jsk-ros-pkg/jsk_robot/blob/93919243019ae06be43874022062abbeee00349e/jsk_panda_robot/panda_eus/euslisp/dual_panda-interface.l, so we have to re-use code in some way.
I'll make a PR for this after the other review comments are resolved or replied.

jsk_panda_robot/panda_eus/models/panda.yaml
```diff
@@ -0,0 +1,19 @@
...
+larm-end-coords:
+  parent: panda_link8
+  translate: [0, 0, 0.1]
+  rotate : [-0.35740972270209, -0.86285515644477, -0.35740630816298, 98.42112728719206]
+
+angle-vector:
+  reset-pose: [ 0.0, -45.0, 0.0, -135.0, 0.0, 90.0, 45.0,
+                0.0, -75.0 ]
+  reset-manip-pose: [ 0.0, -45.0, 0.0, -135.0, 0.0, 90.0, 45.0,
+                      0.0, -75.0 ]
```
pazeshun, line 19: `reset-pose` seems different from https://github.com/jsk-ros-pkg/jsk_robot/blob/93919243019ae06be43874022062abbeee00349e/jsk_panda_robot/panda_eus/models/dual_panda.yaml#L32-L34
Also, `reset-pose` and `reset-manip-pose` are the same.
Do you have any reason why you choose this angle-vector as `reset-pose` and `reset-manip-pose`?

pazeshun (changes requested):

jsk_panda_robot/jsk_panda_startup/launch/panda_control.launch
```diff
@@ -0,0 +1,32 @@
...
+    <arg name="robot_ip" value="$(arg robot_ip)" />
+    <arg name="arm_id"   value="$(arg arm_id)" />
+  </include>
+
+  <node name="franka_control" pkg="franka_control" type="franka_control_node" output="screen" required="true">
+    <rosparam command="load" file="$(find franka_control)/config/franka_control_node.yaml" subst_value="true" />
+    <param name="robot_ip" value="$(arg robot_ip)" />
+  </node>
+  <rosparam command="load" file="$(find franka_control)/config/default_controllers.yaml" subst_value="true" />
+  <node name="state_controller_spawner" pkg="controller_manager" type="spawner" respawn="false" output="screen" args="franka_state_controller position_joint_trajectory_controller" />  <!-- add controller -->
```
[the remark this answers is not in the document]
yuki-asano: In addition to add `position_joint_trajectory_controller` , change the path of xacro to load from `panda_eus` dir,  and add default IP. (comments for modified parts are left in the launch file)

I thought reuse of the code as you mentioned, but I'm not sure it will be correctly overridden. So, I copied the `franka_control.launch` which is not so much volume and left comments.
In order to separate core and parameters, we might need to make something like `franka_robot.launch` based on `franka_control.launch`, and then `panda.launch` can be described with only parameters  of `franka_robot.launch`, such as ip, xacro, controllers etc..

It is ok to change the name as following your preferences. `panda_control.launch` just came from `franka_control.launch`.
[dfa63af](https://github.com/jsk-ros-pkg/jsk_robot/pull/1799/commits/dfa63af836818f3ea5bd52a5cee6b9459e6b0506)

jsk_panda_robot/panda_eus/CMakeLists.txt
yuki-asano: move the code for `panda.l` into `if` block and modify messages.
[7a093af](https://github.com/jsk-ros-pkg/jsk_robot/pull/1799/commits/7a093afe919bc2038fde6d3611f7cf2fba24eed9)

jsk_panda_robot/panda_eus/euslisp/panda-interface.l
yuki-asano: remove this
[d19204b](https://github.com/jsk-ros-pkg/jsk_robot/pull/1799/commits/d19204ba8670ad0b08cdff432a81638453e71bbe)

jsk_panda_robot/panda_eus/euslisp/panda-interface.l
yuki-asano: That's great idea to extract common parts from the two interfaces.
I tried to think about it but it seemed difficult than I thought because a specific robot-model is required for the robot-interface of it here.
https://github.com/jsk-ros-pkg/jsk_robot/pull/1799/files/c943a29e24f7d8175393321b7f215e91d0b69829#diff-49407bb0669ccb3405b9c922a92547506aea3393e03e7c9aa2615efe645ed989R18

I hope to hear how this will be solved.

jsk_panda_robot/jsk_panda_startup/launch/panda_control.launch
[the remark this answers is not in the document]
pazeshun: > change the path of xacro to load from panda_eus dir, and add default IP. (comments for modified parts are left in the launch file)

Could you try the following?
```
<launch>
  <arg name="robot_ip" default="192.168.0.2"/>
  <!-- Specify other arguments here -->

  <include file="$(find franka_control)/launch/franka_control.launch>
    <arg name="robot_ip" value="$(arg robot_ip)" />
    <!-- Specify other arguments here -->
  </include>
  <param name="robot_description" command="$(find xacro)/xacro $(find panda_eus)/models/panda.urdf.xacro hand:=$(arg load_gripper) arm_id:=$(arg arm_id) $(arg xacro_args)"/>

  <node name="position_joint_trajectory_controller_spawner" pkg="controller_manager" type="spawner" respawn="false" output="screen" args="position_joint_trajectory_controller" />

</launch>
```
Maybe you can overwrite `robot_description` by setting it after including `franka_control.launch`.
See https://answers.ros.org/question/199608/roslaunch-order-of-rosparams/ for details.

If this does not work, `panda.launch` without including `franka_control.launch` is OK, but at least, `state_controller_spawner` is a wrong name because it spawns not only `franka_state_controller` but also `position_joint_trajectory_controller`.
I recommend to add `position_joint_trajectory_controller_spawner` separately.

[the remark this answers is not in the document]
yuki-asano: I tried and the overwrite of parameters seems working fine.
https://github.com/jsk-ros-pkg/jsk_robot/pull/1799/commits/3a79e150c8f174dcc34241ba5c556359c52a27be

Thank you for the link and tips of rosparam!

jsk_panda_robot/panda_eus/euslisp/panda-interface.l
pazeshun: @yuki-asano Currently I am working on this and want to ask you a question.
You set limb name as `:larm`:
https://github.com/jsk-ros-pkg/jsk_robot/pull/1799/files#diff-75f2fe8730d1b6c5ca7bf942abc99bf5955664a80692a4e7263428b14268a1c2R1
https://github.com/jsk-ros-pkg/jsk_robot/pull/1799/files#diff-75f2fe8730d1b6c5ca7bf942abc99bf5955664a80692a4e7263428b14268a1c2R10
But in euslisp common sense, single arm should be `:rarm`:
https://github.com/jsk-ros-pkg/jsk_robot/blob/93919243019ae06be43874022062abbeee00349e/jsk_denso_robot/cobottaeus/cobotta.yaml#L5
https://github.com/jsk-ros-pkg/jsk_robot/blob/93919243019ae06be43874022062abbeee00349e/jsk_fetch_robot/fetcheus/fetch.yaml#L7
https://github.com/jsk-ros-pkg/jsk_robot/blob/93919243019ae06be43874022062abbeee00349e/jsk_kinova_robot/kinovaeus/gen3_lite_gen3_lite_2f.yaml#L5

Can I take this opportunity to change the name to `:rarm`?
I'm afraid your code calls `:larm` many times.

yuki-asano: Thank you for telling the common sense of euslisp coding. I will check it out.

yuki-asano: change `:larm` to `:rarm`
[74fc6e4](https://github.com/jsk-ros-pkg/jsk_robot/pull/1799/commits/74fc6e4a871e547dc3855c0f20ab4c0f2ec66c6a)

pazeshun: @yuki-asano Sorry for late, I completed
> extract common parts from the two interfaces

Could you check https://github.com/yuki-asano/jsk_robot/pull/1?

jsk_panda_robot/panda_eus/models/panda.yaml
pazeshun: In https://github.com/yuki-asano/jsk_robot/pull/1, I added notice about the difference of `reset-pose` and `reset-manip-pose` (and also `end-coords`) to README

yuki-asano:
@pazeshun chould you approve changes?

jsk_panda_robot/panda_eus/euslisp/panda-interface.l
pazeshun: Finally resolved by https://github.com/yuki-asano/jsk_robot/pull/2

pazeshun:
> @pazeshun chould you approve changes?

I should not approve until CI passes (https://github.com/jsk-ros-pkg/jsk_robot/pull/1208#issuecomment-1135488176).
And currently, CI seems to fail due to `apt` server problem:
```
  + sudo apt-get install -y --force-yes -q -qq python-rosdep python-wstool python-catkin-tools
  W: --force-yes is deprecated, use one of the options starting with --allow instead.
  E: Package 'python-rosdep' has no installation candidate
  E: Package 'python-wstool' has no installation candidate
  E: Package 'python-catkin-tools' has no installation candidate
```
https://github.com/jsk-ros-pkg/jsk_robot/actions/runs/5150618620/jobs/9276407913?pr=1799
In this situation, we should wait for some time while sometimes restarting CI (too frequent restarting has a negative influence, though).
I'll take care about CI, so please wait.

yuki-asano:
OK. I missed the failed CI in the last. It is hidden and appears by scroling..

pazeshun (approved):
@k-okada 
I think this PR is now OK, but I have one concern.
Currently, `reset-pose`, `reset-manip-pose`, and `end-coords` of `panda` are different from `dual_panda`:
https://github.com/jsk-ros-pkg/jsk_robot/pull/1799/files#diff-75f2fe8730d1b6c5ca7bf942abc99bf5955664a80692a4e7263428b14268a1c2R10-R19
https://github.com/jsk-ros-pkg/jsk_robot/blob/a4b99e4a54afde3df80159435a7db4b90ea45627/jsk_panda_robot/panda_eus/models/dual_panda.yaml#L22-L37
Especially as for `end-coords`, I think `panda`'s `end-coords` is better because it is on the center of the fingertip.
However, we already have some demo codes depending on `dual_panda`'s `end-coords`, so it is not easy to change it.
I think it is OK to leave them as they are because I wrote a notice about them:
https://github.com/jsk-ros-pkg/jsk_robot/pull/1799/files#diff-e424322edffb86d671056c36f1dd1502df5529e99aedfd12f7cb3c101be5f36eR110
What do you think?

pazeshun:
@k-okada Kindly ping

k-okada:
@pazeshun sorry for being late, 

> However, we already have some demo codes depending on dual_panda's end-coords, so it is not easy to change it.

which demo uses this feature? I think we want to change the end coords near future(1-3 years), so how about add warning message to source code too,
```
(:end-coords (&rest args)
  (warning-message 2?3? (yeallow) "the endo coords of dual_pand is differnet from single panda and might be change in near future.)
  (send-super* :end-coords args))
```
in 
https://github.com/jsk-ros-pkg/jsk_robot/pull/1799/files#diff-da9673f8d3480112a7dbd60e430a179c2df50bf7f12437ae7419bd9777e0f357

pazeshun:
> which demo uses this feature?

For example, https://github.com/jsk-ros-pkg/jsk_demos/pull/1363/files#diff-e69875d71091c4531bb74bf07af5b703bc585828d41f8fea5aa8fbde9f4ef1e0R22.
Also I used `end-coords` directly in my previous demo: https://gitlab.jsk.imi.i.u-tokyo.ac.jp/hasegawa/hasegawa_moonshot/-/blob/7b71edf5a7cb1c6ada46214f5aafb8d03f9d6cfa/hasegawa_moonshot/euslisp/lib/drilling-interface.l#L64-L66.
Perhaps @haraduka may also use `end-coords` directly.

> I think we want to change the end coords near future(1-3 years)

If you might allow this, then I think we should change `end-coords` right now, otherwise we will break too many demos because the number of demos will increase.
But just now, we do not have enough time to test previous demos with the new `end-coords`.
Do you OK to keep this PR open until testing finishes?
Or should we just add a warning message and merge this PR?
(But I'm afraid too many warning messages may be shown because `:end-coords` may be called everywhere. Also I may forget to fix `end-coords`...)

yuki-asano:
+1
> Or should we just add a warning message and merge this PR?

jsk_panda_robot/panda_eus/models/panda.yaml
pazeshun: After https://github.com/yuki-asano/jsk_robot/pull/3, `pannda` and `dual_panda` use the same `end-coords`.

pazeshun (approved):
@k-okada 
Now `panda` and `dual_panda` use the same `end-coords`:
https://github.com/yuki-asano/jsk_robot/pull/3#issuecomment-1664954572
Could you review this?

@haraduka and I confirmed that slightly-modified version of @haraduka 's demo code works with this PR.
I'll fix my old demo to work with this PR (it is easy because only small translation is needed).
https://github.com/jsk-ros-pkg/jsk_demos/pull/1363 will also be affected and it is difficult to check because we need to move panda to old pedestals etc., so I just wrote a comment:
https://github.com/jsk-ros-pkg/jsk_demos/pull/1363#issuecomment-1759428040

pazeshun (changes requested):
Sorry, this PR may break gripper movement methods.
I will check deeply and make a PR to fix.

pazeshun (approved):
@k-okada 
Now gripper movement methods were fixed: https://github.com/yuki-asano/jsk_robot/pull/4#issue-1944482368
Could you review this?

Why we could not find that bug at https://github.com/jsk-ros-pkg/jsk_robot/pull/1799#pullrequestreview-1673938413 is that we considered that strange gripper movement as a bug of gripper itself, which we have sometimes faced so far.

k-okada:
@pazeshun LGTM, as for https://github.com/yuki-asano/jsk_robot/pull/1#discussion_r1212576855, can we fix by
https://github.com/jsk-ros-pkg/jsk_roseus/pull/746 ??

pazeshun:
@k-okada Thank you, I confirmed https://github.com/jsk-ros-pkg/jsk_roseus/pull/746 fixes https://github.com/yuki-asano/jsk_robot/pull/1#discussion_r1212576855 even when `:wait nil`:
```
3.irteusgl$ send *ri* :get-start-grasp-result :rarm :wait nil
[ERROR] [1698643393.384314525]: [/franka_gripper/grasp] :get-result called without sending goal, returns null result
nil
```
Can we merge this PR as it is and consider https://github.com/jsk-ros-pkg/jsk_roseus/pull/746 as further improvement?

yuki-asano:
Thank you!! @pazeshun @k-okada

PR #1803, opened 2023-06-07
[jsk_robot_startup/lifelog] Tweet randomly from uptime

tkmtnt7000:
develop/fetch version 

This PR enables robot to tweet for random minutes instead of elapsed time since robot starts.

### DIFF
--- a/jsk_robot_common/jsk_robot_startup/lifelog/tweet_client_uptime.l
+++ b/jsk_robot_common/jsk_robot_startup/lifelog/tweet_client_uptime.l
@@ -19,6 +19,13 @@
   (ros::ros-info "Wait for /active_user/elapsed_time parameter ..."))
 
 (cond
+ ((ros::has-param "/active_user/tweet_random_range")
+  (setq *tweet-random-range* (ros::get-param "/active_user/tweet_random_range")))
+ (t
+  (setq *tweet-random-range* 600.0)
+  ))
+
+(cond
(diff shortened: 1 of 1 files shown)


k-okada:
@tkmtnt7000 I think this feature could be the default.

tkmtnt7000:
Yes, I would like to execute this node in the default.

k-okada:
I mean enable this feature withtout explictly define `"/active_user/tweet_random_range"` in each launch files.

jsk_robot_common/jsk_robot_startup/lifelog/tweet_client_uptime.l
```diff
@@ -18,6 +18,13 @@
   (unix::sleep 3)
   (ros::ros-info "Wait for /active_user/elapsed_time parameter ..."))
 
+(cond
+ ((ros::has-param "/active_user/tweet_random_range")
+  (setq *tweet-random-range* (ros::get-param "/active_user/tweet_random_range")))
+ (t
+  (setq *tweet-random-range* 600.0)
```
tkmtnt7000, line 25: I think this section sets random range without explicitly setting rosparm `/active_user/tweet_random_range`

k-okada: I see, you're right.

PR #1804, opened 2023-06-07
[Related #1725] [jsk_robot_startup/lifelog] Tweet randomly from uptime

tkmtnt7000:
master version of #1803 

I think #1725 is important to control tweet voice. (I forcefully resolved conflicts in cherry-picking #1725's commit)


jsk_robot_common/jsk_robot_startup/lifelog/tweet_client_tablet.l
```diff
@@ -3,13 +3,42 @@
 (ros::roseus "twitter_client_tablet")
 (ros::roseus-add-msgs "roseus")
 
+(load "package://roseus/euslisp/dynamic-reconfigure-server.l")
```
k-okada, line 6: this code does not work without https://github.com/jsk-ros-pkg/jsk_roseus/pull/729, so please add comment, 
it is also batter to 
```
(if (not (probe-file "...dynamic-reconfigure-server.l"))
  ...
)
```
lik https://github.com/jsk-ros-pkg/jsk_pr2eus/blob/8fc634d0d92830c201f9ae922c5e673c0def0b19/pr2eus/make-pr2-model-file.l#L1, but if it is too long, maybe manually write 
```(ros::advertise-service service-name dynamic_reconfigure::Reconfigure                              #'send self :set-callback)```
woud be ok.

PR #1815, opened 2023-06-18
Add labeler action for develop/fetch

sktometometo:
Add https://github.com/actions/labeler in order to add `develop/fetch` label to pull requests for `develop/fetch` branch.

See https://github.com/jsk-ros-pkg/jsk_robot/pull/1814 to check how it works.


tkmtnt7000 (approved):
Thank you very much for setting nice feature!

It's ok to merge them into the develop/fetch branch as they are.
It would be appreciated if you could combine the related commits into one when you send PR to the master branch.

PR #1816, opened 2023-06-18
Add labeler action

sktometometo:
Add https://github.com/actions/labeler in order to add labels for each robot based on changed directories.

See https://github.com/jsk-ros-pkg/jsk_robot/pull/1814 to check how it works.


tkmtnt7000 (approved):
Superb!

PR #1825, opened 2023-06-22
Add dstat supervisor job for Fetch

sktometometo:


sktometometo:
@k-okada

sktometometo:
@708yamaguchi Could you explain what is it and why do we need it?

708yamaguchi:
I was monitoring fetch's computer resources to investigate a fetch freeze problem.

PR #1828, opened 2023-06-27
[WIP][jsk_fetch_startup][jsk_fetch_robot] add esp_now_ros to launch

sktometometo:
Use https://github.com/sktometometo/esp_now_ros in fetch1075 by default

### DIFF
--- a/jsk_fetch_robot/jsk_fetch_startup/launch/fetch_bringup.launch
+++ b/jsk_fetch_robot/jsk_fetch_startup/launch/fetch_bringup.launch
@@ -47,6 +47,7 @@
   <!--iot devices -->
   <arg name="launch_rosserial" default="true" />
   <arg name="launch_switchbot" default="true" />
+  <arg name="smart_device_port" default="$(optenv SMART_DEVICE_PORT)"/>
 
   <!-- lifelog -->
   <arg name="launch_fetch_lifelog" default="true" />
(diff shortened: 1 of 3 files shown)


jsk_fetch_robot/jsk_fetch_startup/launch/fetch_bringup.launch
```diff
@@ -422,6 +423,13 @@
     </group>
   </group>
 
+  <!-- smart device ros -->
+  <group if="$(eval arg('smart_device_port') != '')">
```
tkmtnt7000, line 427: I'm just commenting because there are no robots running on indigo at JSK right now, so I'm sure it's fine, but I've received comments before that this writing style is deprecated because it is only supported by kinetic and later.

tkmtnt7000: Here https://github.com/jsk-ros-pkg/jsk_robot/pull/1720#discussion_r1046673452.

tkmtnt7000: But I think it's ok to this PR because that #1720 PR is for jsk_robot_startup.

sktometometo: I see, thanks for your comments.

PR #1798, opened 2023-05-07
[jsk_aibo_robot][aibo_selenium_ros] add package

sktometometo:
Add aibo_selenium_ros

PR #1852, opened 2023-09-15
PR/add jsk aibo robot

a-ichikura:
we add `version_main` parameter into __init__.py because the version of undetected driver should be as same as chrome-driver and chrome-browser.


jsk_aibo_robot/aibo_selenium_ros/python/aibo_selenium_ros/geckodriver.log
```diff
@@ -0,0 +1,71 @@
+1694083040310	geckodriver	INFO	Listening on 127.0.0.1:36887
```
sktometometo, line 1: In order to keep the code base minimal, files that are generated later, such as log files, generally speaking, should not be included in the git repository.

sktometometo: @a-ichikura Could you update `jsk_aibo_robot/aibo_selenium_ros/.gitignore` so that this file will not be added to git repo?

jsk_aibo_robot/aibo_selenium_ros/python/aibo_selenium_ros/__init__.py
```diff
@@ -0,0 +1,179 @@
...
+
+    self.initialized = False
+
+    options = Options()
+    if headless:
+      options.add_argument('--headless')
+    if chrome_executable_path is not None:
+      options.binary_location = chrome_executable_path
+
+    self.driver = uc.Chrome(executable_path=webdriver_path, options=options, version_main=115)
```
sktometometo, line 36: chrominum version is different for each system. So it will be convenient if this program retrieve main version from `chromium-browser --version` command. See https://github.com/jsk-ros-pkg/jsk_robot/pull/1798/commits/7cbcd36e9480f871469fd1adafe0a1e39e914218 for reference.

sktometometo (changes requested):

PR #1856, opened 2023-09-24
[jsk_fetch_robot] Downgrading moveit_msgs in noetic client PC for sending angle-vector

mqcmd196:
fixes https://github.com/jsk-ros-pkg/jsk_robot/issues/1855

### DIFF
--- a/jsk_fetch_robot/README.md
+++ b/jsk_fetch_robot/README.md
@@ -43,7 +43,7 @@ mkdir -p catkin_ws/src
 cd  catkin_ws/src
 wstool init .
 wstool set --git jsk-ros-pkg/jsk_robot https://github.com/jsk-ros-pkg/jsk_robot.git -y
-if [[ $ROS_DISTRO =~ ^(indigo|kinetic|melodic)$ ]]; then
+if [[ $ROS_DISTRO =~ ^(indigo|kinetic|melodic|noetic)$ ]]; then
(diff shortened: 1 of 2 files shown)


k-okada:
Please consider server-side (robot-side) solution, instead of requireing each users to setup individually, specially this solution requires different workspace for different robots, that means, if you prepared a demo code for your robot, and try to apply that system to other robots, you need to setup different worksapce. 

If the differentce of `moveit_msgs` are CollisionObjects.msg (https://github.com/ros-planning/moveit_msgs/compare/0.11.4...melodic-devel) and that is only used to publish pr2eus to moveit, one ides it to start node that take noetic veresion of CollisionObjects within robot, as AnyMsg and republish as CollisionMsg as original topic name (https://github.com/jsk-ros-pkg/jsk_robot/pull/898), If you need to subscribe melodic version of CollisionObject within roseus running at noetic client , you can use https://github.com/jsk-ros-pkg/jsk_roseus/pull/631

mqcmd196:
I agree with your point. Close then create another PR

mqcmd196:
BTW, we can use https://github.com/jsk-ros-pkg/jsk_robot/pull/898 on topic, but do you know any options to do it on service?

mqcmd196:
Hmm, create rospy.AnySrv

PR #1873, opened 2023-10-13
[Unitree][cross] set make job num to CPU + 1

sktometometo:
Make job number of make to CPU + 1 to make it adaptive


sktometometo:
Got some error, revert to draft

Issue #1870, opened 2023-10-11
[Unitree] install.sh does not work


sktometometo:
Install.sh does not work.

- PC: Thinkpad P15s
- OS: Ubuntu 20.04
- ROS Noetic

git diff to master

```diff
[12:20 午前 木 10月 12]
~/ros/ws_jsk_unitree/src/jsk-ros-pkg/jsk_robot/jsk_unitree_robot/cross $ git diff master
diff --git a/jsk_unitree_robot/cross/repos/unitree.repos b/jsk_unitree_robot/cross/repos/unitree.repos
index 172aae74f..4e83970d1 100644
--- a/jsk_unitree_robot/cross/repos/unitree.repos
+++ b/jsk_unitree_robot/cross/repos/unitree.repos
@@ -24,3 +24,15 @@ repositories:
     # enable to run simple apps parallel to other apps
     # https://github.com/PR2/app_manager/pull/59 have been merged,
     # but 1.4.0 have not released yet.
+  esp_now_ros:
+    type: git
+    url: https://github.com/sktometometo/esp_now_ros
+    version: master
+  rosserial:
+    type: git
+    url: https://github.com/ros-drivers/rosserial
+    version: melodic-devel
+  catkin_virtualenv:
+    type: git
+    url: https://github.com/locusrobotics/catkin_virtualenv.git
+    version: master
diff --git a/jsk_unitree_robot/jsk_unitree_startup/package.xml b/jsk_unitree_robot/jsk_unitree_startup/package.xml
index 9723686b9..85056213c 100644
--- a/jsk_unitree_robot/jsk_unitree_startup/package.xml
+++ b/jsk_unitree_robot/jsk_unitree_startup/package.xml
@@ -16,6 +16,7 @@
   <run_depend>rosserial_python</run_depend>
   <run_depend>teleop_twist_joy</run_depend>
   <run_depend>respeaker_ros</run_depend>
+  <run_depend>esp_now_ros</run_depend>
 
   <!-- for diagnostic aggregator -->
   <run_depend>diagnostic_aggregator</run_depend>
sktometometo@lee
[gitbranch:experiment_uwb_localization_master]
[ROS VERSION:1]

```

Error log

```bash
~/ros/ws_jsk_unitree/src/jsk-ros-pkg/jsk_robot/jsk_unitree_robot/cross $ ifconfig enp0s31f6
[...]

PR #1874, opened 2023-10-15
[cross] add LANG=C for ping reachability check

sktometometo:
Fix https://github.com/jsk-ros-pkg/jsk_robot/issues/1870


k-okada:
LANG=C is always good

PR #1885, opened 2023-10-31
add single fr3

yuki-asano:
- added minimum set of fr3 (franka research 3) model which is almost same with panda model (only name is different)
- confirmed motions of the real robot by

```
(send *ri* :angle-vector (send *fr3* :angle-vector) 5000)
(send *ri* :start-grasp :rarm)
(send *ri* :stop-grasp :rarm)
```

Screenshot from 2023-10-31 12-31-44

### DIFF
--- a/jsk_panda_robot/README.md
+++ b/jsk_panda_robot/README.md
@@ -49,17 +49,17 @@
    ```
 
 
-## Running single Panda
+## Running single Panda/FR3 (Franka Research 3)
 ### Boot robot
 1. Please turn on the controller box and unlock joints by accessing desk.
 ### Via roseus
 1. Start controller on controller PC:
    ```bash
    ssh leus@dual-panda1.jsk.imi.i.u-tokyo.ac.jp  # Or ssh leus@dual-panda2.jsk.imi.i.u-tokyo.ac.jp
-   roslaunch jsk_panda_startup panda.launch robot_ip:=<IP OF TARGET ROBOT>
+   roslaunch jsk_panda_startup panda.launch robot_ip:=<IP OF TARGET ROBOT>  # FR3: roslaunch jsk_panda_startup fr3.launch robot_ip:=<IP OF TARGET ROBOT>
    ```
 
-2. Controlling single Panda via roseus:
+2. Controlling single Panda/FR3 via roseus:
    1. Setting up network:
       ```bash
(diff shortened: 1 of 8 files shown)


pazeshun (approved):
JSK does not have real fr3, so I only confirmed that `fr3.l` seems correctly generated and the following procedure works on kinematics simulator:
```lisp
(fr3-init)
(objects (list *fr3*))
(send *ri* :stop-grasp :rarm)
(send *ri* :start-grasp :rarm)
(send *fr3* :stop-grasp :rarm)
(send *fr3* :start-grasp :rarm)
(send *fr3* :init-pose)
(send *ri* :angle-vector (send *fr3* :angle-vector) 5000)
```

k-okada:
if the only difference is name , why not load the existing lisp  file and
change robot name?
--
◉ Kei Okada



2023年10月31日(火) 20:45 Shun Hasegawa ***@***.***>:

pazeshun:
@k-okada 

> if the only difference is name , why not load the existing lisp file and change robot name?

I think @yuki-asano 's explanation is incorrect.

At least, joint names are different, so we have to generate `fr3.l`:
https://github.com/jsk-ros-pkg/jsk_robot/pull/1885/files#diff-3518a932c47d029d01ebd0f0d3e9f8d0656adbca5d2c218f48b9e326124596e1R2-R8
vs.
https://github.com/jsk-ros-pkg/jsk_robot/blob/ddf454dd5f4f4c1aa43b81fda6e17a24d96c721f/jsk_panda_robot/panda_eus/models/panda.yaml#L2-L8

So it requires a big change to reuse `panda-utils.l` in `fr3-utils.l`, because current `panda-utils.l` uses `defmethod` against `panda-robot`, not `fr3-robot`:
https://github.com/jsk-ros-pkg/jsk_robot/blob/ddf454dd5f4f4c1aa43b81fda6e17a24d96c721f/jsk_panda_robot/panda_eus/euslisp/panda-utils.l

As for `fr3-interface.l`, we may be able to reuse `panda-interface.l` by defining `fr3-interface` as a child class of `panda-interface`. I do not think this is a good idea because code uniformity (parent class of all franka robots' interfaces is `franka-common-interface`) becomes lost, but I'll follow @k-okada 's decision.

pazeshun (approved):
@k-okada Maybe I followed your comment. Could you check this PR again?

pazeshun:
@k-okada Kindly ping

pazeshun:
@k-okada Kindly ping

pazeshun:
@k-okada Kindly ping

pazeshun:
@k-okada Kindly ping

PR #1896, opened 2023-12-05
[Bugfix] [jsk_pr2_startup] Remove a slash of the first character for noetic (tf2) in publish_empty_cloud.py

tkmtnt7000:
This PR removed slashes from the first character of the topics for noetic tf2 compatibility.

Without this PR, tf is not solved and PR2 does not move with `move_base`  when we use `empty_cloud` in navigation that PR2 Subway demostration uses.

cc: Thanks to @pazeshun


tkmtnt7000:
With this PR, we can go inside the elevator in subway demonstration!

mqcmd196 (approved):
Tested on the real robot with subway demo. Thanks a lot! LGTM

PR #1914, opened 2024-06-03
add args to switch d435 and l515 as realsense_torso camera  & change the extrinsic parameters of them

soonhyo:
I add camera_type args in launches and changed the extrinsic parameters of them.

### DIFF
--- a/jsk_baxter_robot/jsk_baxter_startup/jsk_baxter_lifelog/estimated_endeffector.l
+++ b/jsk_baxter_robot/jsk_baxter_startup/jsk_baxter_lifelog/estimated_endeffector.l
@@ -5,22 +5,26 @@
 
 (if (not (boundp '*baxter*))
     (setq *baxter* (instance baxter-robot :init)))
+(setq frames (make-hash-table))
+(dolist (arm (list :rarm :larm))
+  (sethash arm frames (send (send (send *baxter* arm :end-coords) :parent) :name))
+  )
 
-(ros::rate 20)
+(ros::rate 100)
 
 (defun right-state-cb(msg)
-  (state-cb "/robot/end_effector/right/wrench" :rarm (send msg :wrench))
+  (state-cb "/robot/end_effector/right/wrench" :rarm (send msg :wrench) (send msg :header :stamp))
   )
 
 (defun left-state-cb(msg)
-  (state-cb "/robot/end_effector/left/wrench" :larm (send msg :wrench))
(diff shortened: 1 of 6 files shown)


jsk_baxter_robot/jsk_baxter_startup/jsk_baxter_lifelog/estimated_endeffector.l
```diff
@@ -5,27 +5,31 @@
 
 (if (not (boundp '*baxter*))
     (setq *baxter* (instance baxter-robot :init)))
+(setq frames (make-hash-table))
+(dolist (arm (list :rarm :larm))
+  (sethash arm frames (send (send (send *baxter* arm :end-coords) :parent) :name))
+  )
 
-(ros::rate 20)
+(ros::rate 100)
```
iory, line 13: I think this is a good PR. As a comment, if you include the reason for changing the loop rate in the code, it will help readers understand why this value was chosen.

iory (changes requested):

mqcmd196 (commented):
Why do you want to switch but not to launch both? The old camera was removed?

jsk_baxter_robot/jsk_baxter_startup/jsk_baxter_lifelog/estimated_endeffector.l
soonhyo: the 100 Hz is the same rate of Baxter's joint state. This is why I set the rate into 100 Hz.  
But I didn't mean to PR with this commit... It is just my mistake.
Is it better to separate the camera setting PR and end_effector PR?

iory: I see.
In that case, it's better to split the PR. Adding an explanation in the EusLisp code about the rate changes in the separated PR would be good.

soonhyo:
> Why do you want to switch but not to launch both? The old camera was removed?

Nope, there are two cameras on Baxter. But now there are no user using both of them simultaneously.

jsk_baxter_robot/jsk_baxter_startup/jsk_baxter_lifelog/estimated_endeffector.l
soonhyo: Okay, thanks you.
I will PR this changes again.

PR #1926, opened 2024-12-04
[jsk_pr2_startup] respawn speech_to_text

mqcmd196:


Kanazawanaoaki (approved):
In my own testing, I didn't see the phenomenon where `speech_to_text` dies after `respeaker_node` dies, but I think it's fine since {speech_to_text{ respawns normally.

[log]( https://gist.github.com/Kanazawanaoaki/498d4cc720ce3004dff462d9326b4fba#file-20241209_respeaker_test-log)


https://github.com/user-attachments/assets/63480c72-88b2-4cfa-9b70-b61017b7c4d4


While pinging the two nodes, I first physically unplugged the respeaker and observed the progress.
The `speech_to_text` node did not die naturally, so I killed it and confirmed that it respawned.
I plugged the respeaker back in and confirmed that speech recognition was performed.

PR #1929, opened 2024-12-16
[jsk_pr2_startup] Add cable connector check warning

Kanazawanaoaki:
I added a warning about the power cable connector, which has recently been causing problems, such as https://github.com/jsk-ros-pkg/jsk_robot/issues/1644.  

We found that there is a correlation between the temperature of the connector and the temperature of the battery, and that the connector begins to melt at around 45℃, so we have added a warning that is triggered by the battery temperature.

### DIFF
--- a/jsk_pr2_robot/jsk_pr2_startup/jsk_pr2_warning/battery_warning.py
+++ b/jsk_pr2_robot/jsk_pr2_startup/jsk_pr2_warning/battery_warning.py
@@ -70,20 +70,6 @@ class BatteryWarning(object):
         self.speak_client.send_goal(SoundRequestGoal(sound_request=req))
         self.speak_client.wait_for_result(timeout=rospy.Duration(10))
 
-    def play_sound(self, sound):
-        key = sound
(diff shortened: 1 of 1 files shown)


jsk_pr2_robot/jsk_pr2_startup/jsk_pr2_warning/battery_warning.py
```diff
@@ -66,6 +70,20 @@ def speak(self, sentence):
         self.speak_client.send_goal(SoundRequestGoal(sound_request=req))
         self.speak_client.wait_for_result(timeout=rospy.Duration(10))
 
+    def play_sound(self, sound):
```
mqcmd196, line 73: `play_sound` function is only used in a single place. I prefer not to create the function and write the whole code directly where it is used.

Kanazawanaoaki: You are right, so I've made the correction.

mqcmd196 (approved):
LGTM

PR #1935, opened 2025-03-10
[jsk_pr2_startup]  Disable cable connector warning

Kanazawanaoaki:
This PR eliminates warnings of the power cable connector added by https://github.com/jsk-ros-pkg/jsk_robot/pull/1929.

This warning is no longer necessary since the cable connector issue has been resolved by the repair in https://github.com/jsk-ros-pkg/jsk_robot/issues/1644#issuecomment-2709574624 .


mqcmd196 (approved):
Now, PR1040's cable and socket are replaced. We don't need the original patch anymore. LGTM

